PR #50, opened 2015-01-03
adapted to master changes

flying-sheep:
dunno if changing `x.to_f64().unwrap()` to `x as f64` is right, but the tests still pass…


BurntSushi:
Thank you!

flying-sheep:
np!

PR #54, opened 2015-01-09
update to rust nightly

japaric:
- uint/int -> usize/isize
- `box foo` -> `Box::new(foo)`
- `hash` module and `String/Show` fallout
- added `#[allow(unstable)]` and `#[feature(int_uint)]` to avoid a flood of warnings (but you may want to audit their use in a follow up commit)


BurntSushi:
Thanks!

PR #68, opened 2015-03-08
Update to Rust master

apasel422:


BurntSushi:
Interesting. Thanks! And thanks for re-enabling debug info!

BurntSushi:
I've uploaded this fix to crates.io in `0.2.3`.

PR #70, opened 2015-03-12
Extend the sieve example slightly

jakub-:
This PR extends the sieve example to advocate constructing properties that are as strict as possible. Currently the `prop_all_prime` only tests that the numbers returned by the sieve are all prime but it doesn't ensure the list is complete. I thus extended `README.md` with another property that tests the sieve for equivalence with a naive primality test.


BurntSushi:
This is a phenomenal fix. Thank you so much! :D

PR #71, opened 2015-03-19
update to Rust master

apasel422:
I didn't replace the uses of the deprecated `range_step` because that function's misuse is the point of the tests that use it. However, as it will be removed eventually, we should find a better example.

### DIFF
--- a/examples/out_of_bounds.rs
+++ b/examples/out_of_bounds.rs
@@ -1,13 +1,10 @@
-#![feature(core)]
-
 extern crate quickcheck;
 
-use std::iter::range;
 use quickcheck::{TestResult, quickcheck};
 
 fn main() {
     fn prop(length: usize, index: usize) -> TestResult {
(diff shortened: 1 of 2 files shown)


BurntSushi:
Thanks! I'm hoping some kind of replacement for `range_step` shows up...

apasel422:
The replacement is the [`step_by` adaptor](http://doc.rust-lang.org/std/ops/struct.Range.html#method.step_by), but it doesn't have the same problem that the example uses as motivation, necessarily.

apasel422:
(Using it as a drop-in replacement works without other modification, but I'm not sure the motivation is there. If it is, I can certainly push an additional commit.)

BurntSushi:
Ah yes, I see. I'll take a look later at the example see what I can come up with. Maybe I'll need to look for a new bug. :-)

BurntSushi:
I've uploaded this fix to crates.io in `0.2.4`.

PR #73, opened 2015-04-04
remove deprecated `PhantomFn` supertrait

apasel422:


BurntSushi:
Hmm, I don't think I can merge this PR as is because it will break `quickcheck` on beta. I haven't really figured out how I want to handle diverging nightly and beta.

One way forward now is to just use `#[allow(deprecated)]`. But this isn't a tenable way forward for all breaking changes.

BurntSushi:
I'm just going to use `allow(deprecated)` for now until I'm forced to do something else. Thanks!

PR #74, opened 2015-04-17
Added lazy shrinking for Vec

### LINKED ISSUE #4: make vector shrinking lazy
Currently, a vector is shrunk by producing all possible shrunk values immediately. It would be much better if they were generated lazily with an iterator.

Perhaps `std::iter::Unfold` will be useful for this.

Marwes:
Adds lazy shrinking for Vec. I have not bench marked this implementation but considering that it does not additional work it should be roughly as fast when the entire iterator is used and significantly faster otherwise.

I added the behavior where the empty vector is added as the first element to the iterator though I could not find this behavior in quickcheck for haskell. Considering how the shrunk vectors actually looks I'd be partial to removing that special case in case I can update this PR.

Fixes #4.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -216,32 +216,92 @@ impl<A: Arbitrary> Arbitrary for Vec<A> {
     }
 
     fn shrink(&self) -> Box<Iterator<Item=Vec<A>>+'static> {
-        if self.len() == 0 {
-            return empty_shrinker();
-        }
+        VecShrinker::new(self.clone())
+    }
+}
 
-        // Start the shrunk values with an empty vector.
-        let mut xs: Vec<Vec<A>> = vec![vec![]];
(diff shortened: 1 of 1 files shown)


BurntSushi:
\o/

This. Is. Awesome!!

BurntSushi:
RE empty vec: I can't remember precisely why I did this, I probably thought "check a common corner case first." I don't have a strong opinion about it, but it does seem like a sensible thing to do unless there are downsides that I'm not considering.

BurntSushi:
I've uploaded this to crates.io in `0.2.14`.

Marwes:
I'm not entirely familiar with the inner workings of quickcheck but running verboseCheck in Haskell it seems that the empty list is naturally generated since it starts by testing small values. Not sure if you do this though.

PR #79, opened 2015-05-19
Implement Testable for more generic results

aepsil0n:
It is sufficient to constrain the error type to be `Debug + Send + 'static` instead of limiting it to `String` only.


BurntSushi:
Love it! Thanks. :-)

PR #80, opened 2015-05-19
Implement Testable for ()

aepsil0n:
This allows for some useful patterns, such as

- testing functions without a return value that only fail by panicking,
- testing functions that return a `Result<(), ...>` that only fail by returning an error.

(This is based on top of #79.)


BurntSushi:
Good idea! Thanks! :-)

PR #81, opened 2015-05-26
Update for rust changes as of 7cb9914fc 2015-05-25

bluss:
ItemFn changed, with this it builds again.

The second commit updates from Modifier to MultiModifer, the former is deprecated.

### DIFF
--- a/quickcheck_macros/src/lib.rs
+++ b/quickcheck_macros/src/lib.rs
@@ -16,7 +16,7 @@ use syntax::ast::Ty_::TyBareFn;
 use syntax::ast_util;
 use syntax::codemap;
 use syntax::parse::token;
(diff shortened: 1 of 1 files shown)


bluss:
Please check the second commit carefully, I don't have experience in syntax extensions :) Just did it to fix the deprecation warning. My testsuite for itertools now compiles again and still seems to work.

I wouldn't have hunted size_hint and fusing bugs without quickcheck, that would be tedious!

BurntSushi:
Thanks! Looks good to me. I'll double check and upload to crates.io.

> I wouldn't have hunted size_hint and fusing bugs without quickcheck, that would be tedious!

^_^

BurntSushi:
OK, updated crates.io to `0.2.19`.

PR #89, opened 2015-08-18
Update README, enable plugin only for testbuild

fkjogu:
When adding the quickcheck dependencies only for dev mode, the plugin also has to be loaded only for testing.

Otherwise the compiler complains about missing dependency of quickcheck_macros.


BurntSushi:
Thanks!

PR #96, opened 2015-09-21
Use `env_logger` to support `RUST_LOG` again

### LINKED ISSUE #95: RUST_LOG has no effect
README.md:

> N.B. When using quickcheck (either directly or via the attributes), RUST_LOG=quickcheck enables info! so that it shows useful output (like the number of tests passed). This is not needed to show witnesses for failures.

`RUST_LOG` no longer controls the output of the `log` crate itself. The old behavior can be replicated with [`env_logger`](https://crates.io/crates/env_logger), or the logging behavior can be changed (and the documentation be updated).

tari:
Debug macros are only used in one place, so logger initialization can simply happen there.

Fixes #95.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,5 +14,6 @@ license = "Unlicense/MIT"
 name = "quickcheck"
 
 [dependencies]
+env_logger = "0.3"
 log = "0.3"
 rand = "0.3"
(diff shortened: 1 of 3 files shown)


BurntSushi:
Oh. That was easy! Thanks!

PR #100, opened 2015-10-13
Add autoshrinking.

SeanRBurton:
I appreciate that this is a very large pull request, and I'm open to breaking it up. This change adds automatic shrinking to `StdGen`, so users no-longer have to implement `shrink` alongside `arbitrary` for their types. The output of the random number generator is cached into a vector, and to shrink testcases we replay simplified versions of this random buffer; this allows us to shrink data structures without knowing anything about them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,5 +16,4 @@ name = "quickcheck"
 [dependencies]
 env_logger = "0.3"
 log = "0.3"
-num = "*"
 rand = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,4 +16,5 @@ name = "quickcheck"
 [dependencies]
 env_logger = "0.3"
 log = "0.3"
+num = "*"
 rand = "0.3"
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -133,6 +133,9 @@ pub fn single_shrinker<A: 'static>(value: A) -> Box<Iterator<Item=A>> {
 /// They must also be sendable and static since every test is run in its own
 /// thread using `thread::Builder::spawn`, which requires the `Send + 'static`
 /// bounds.
+///
+/// If you do not implement `shrink`, StdGen's automatic shrinking feature will
+/// typically do a reasonable job for you.
 pub trait Arbitrary : Clone + Send + 'static {
     fn arbitrary<G: Gen>(g: &mut G) -> Self;
 
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -20,7 +20,7 @@ use rand::Rng;
 ///
 /// A value with type satisfying the `Gen` trait can be constructed with the
 /// `gen` function in this crate.
-pub trait Gen : Rng + Sized {
+pub trait Gen : Rng {
     fn size(&self) -> usize;
     fn shrink_gen(&mut self) -> bool { false }
     fn unshrink_gen(&mut self) { }
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -1,10 +1,4 @@
-use rand::{Rng, Open01, Closed01};
-use rand::distributions::normal::StandardNormal;
-use rand::distributions::exponential::Exp1;
-use std::hash::{Hash, Hasher, SipHasher};
-use std::marker;
-use std::fmt::Debug;
-use std::ops::{Range, RangeFrom, RangeTo, RangeFull};
+use std::char;
 use std::collections::{
     BTreeMap,
     BTreeSet,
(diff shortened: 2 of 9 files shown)


BurntSushi:
Thanks for the PR! The idea seems cool. My major concern is that, in my experience, the art of quickcheck demands one write good shrinkers. Many times, those shrinkers need to take invariants about the data structures into account. For example, with your patch, could one still achieve shrinking like seen here? https://github.com/rust-lang-nursery/regex/blob/master/regex-syntax/src/properties.rs

Could you explain a bit more about what auto shrinking is and how it works? Perhaps the two can co-exist, because I do also admit that many times, writing your own shrinkers can be boiler platey.

SeanRBurton:
Autoshrinking tries to use a variety of strategies to simplify the random numbers used to generate testcases. At the moment I have the following strategies implemented:

Zeroing out blocks of input.

Reducing input words (64, 32, and 8 bit) modulo the generator size. The idea here is that often the arbitrary instance will modulo a random number by the size, so if we do that before-hand we can simplify the random input without changing the generated test-case.

Subtracting constants (64, 32, and 8 bit). We try to subtract fixed constants which approximate powers of two, in order of decreasing size.

Dividing by constants. Similar to above, but we divide instead of subtracting.

Of course, it would be nice to have more strategies, but this is a good start.

SeanRBurton:
As to the regex example, It's a little hard to answer your question in the abstract, but if you can give me an interesting failing test I would be happy to see what happens.

BurntSushi:
It will be a bit before I can get you a concrete example. I'm at work, so I can only deal in the abstract right now.

The problem I have specifically is that shrinking can't always be done generically, so removing the option for the user to write their own shrinker seems a little misguided. For example, in the regex shrinker, it maintains several invariants of the AST while shrinking. For instance, capture indices need to be [correctly ordered](https://github.com/rust-lang-nursery/regex/blob/master/regex-syntax/src/properties.rs#L283). The tree has other invariants, for example a `Concat(Vec<Expr>)` must always contain two or more exprs, otherwise the concat is removed. Similar for `Alternate(Vec<Expr>)`.

Here's another example that shrinks CSV data: https://github.com/BurntSushi/xsv/blob/master/tests/tests.rs#L133 It tries to shrink each individual row, but it must also ensure that every row has the same length.

BurntSushi:
How are `Vec<T>`s shrunk in this scheme?

SeanRBurton:
Everything is shrunk in a uniform way. Say that the StdGen's random number generator emits the following random numbers when generating the test case:
```
1, 3, 7, 9, 13
```
If we then replay these numbers to the arbitrary instance, we should recover the same test-case.

To shrink the test-case, we replay simplified versions of these numbers to the arbitrary instance. For example:
```
1, 3, 7, 0, 0
0, 3, 7, 9, 13
1, 3, 7, 9, 12
```
When we find a sequence of input numbers which generates a failing input, we take that as our new starting point and continue shrinking.

BurntSushi:
I see... that's pretty cool. So, as long as the `arbitrary` method produces correct values, then the shrinker will too. Interesting.

Is there any way to have this and the current more explicit approach?

Have other quickcheck libraries in other languages used this approach before? How has it worked out?

SeanRBurton:
I don't know of any quickcheck libraries using this, but American Fuzzy Lop (AFL) does something very similar. 

We could layer the previous approach on top, and of course people can take the failing input and apply whatever post-processing they want to it, but my hope was that this would make explicit shrinking obsolete.

BurntSushi:
@SeanRBurton OK, I thought about this PR this morning and my feeling is that we would be remiss to not at least try this. It seems wicked cool to me. However, I think it is premature to remove explicit shrinking. I personally having a non-trivial body of code using explicit shrinkers, so for me to try this new feature as is, I'd have to switch wholesale, which doesn't seem like an attractive proposition.

With that said, could you find a way to expose this as a non-default alternative? Perhaps a new method on `QuickCheck` and maybe a new top-level convenience function too. If this works as well as you think it might, *then* we can talk about deprecating explicit shrinkers (with feedback from others who use this library).

Finally, is it feasible to separate out the addition of the associated type to `Testable` and the type parameter to `TestResult`? That actually looks like a good change to me, but I'm not sure if it belongs in this PR. (It would also be nice to be able to land your automatic shrinking without a minor version bump!)

SeanRBurton:
Ok, I've substantially reduced the scope of this change. Everything should now be completely backwards-compatible.

SeanRBurton:
Ping.

Cargo.toml
```diff
@@ -16,4 +16,5 @@ name = "quickcheck"
 [dependencies]
 env_logger = "0.3"
 log = "0.3"
+num = "*"
```
BurntSushi: Could you please remove the `*` dependency on `num`? Wildcards are not good because it subjects downstream users to arbitrary breakage.

src/arbitrary.rs
```diff
@@ -39,21 +46,69 @@ pub struct StdGen<R> {
...
 }
 
 impl<R: Rng> Rng for StdGen<R> {
-    fn next_u32(&mut self) -> u32 { self.rng.next_u32() }
-
-    // some RNGs implement these more efficiently than the default, so
-    // we might as well defer to them.
-    fn next_u64(&mut self) -> u64 { self.rng.next_u64() }
-    fn fill_bytes(&mut self, dest: &mut [u8]) { self.rng.fill_bytes(dest) }
+    #[inline(always)]
```
BurntSushi: Could you remove the `inline(always)` annotations? Unless there's an observable benefit to performance, I'd prefer to leave this up to llvm.

BurntSushi:
This is looking pretty good. I left a couple comments. Once those are addressed, I think I'm ready to merge!

SeanRBurton:
Here you go. :)

SeanRBurton:
Ping.

BurntSushi:
OK, let's do this. I'm going to refrain from making a release until I personally have been able to try it. (I have not had time yet to do so.)

Wilfred:
Note that the author of hypothesis believes that autoshrinking is the right approach rather than explicit shrinkers: http://hypothesis.works/articles/integrated-shrinking/

BurntSushi:
I'm not intrinsically opposed to autoshrinking.

PR #101, opened 2015-10-20
Significant refactoring and simplification.

SeanRBurton:
Does what it says on the tin.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -22,7 +22,6 @@ use rand::Rng;
 /// `gen` function in this crate.
 pub trait Gen : Rng {
     fn size(&self) -> usize;
-    fn reset(&mut self) {}
     fn shrink_gen(&mut self) -> bool { false }
     fn unshrink_gen(&mut self) { }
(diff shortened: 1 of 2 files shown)


src/arbitrary.rs
```diff
@@ -20,6 +20,7 @@ use rand::Rng;
 /// `gen` function in this crate.
 pub trait Gen : Rng {
     fn size(&self) -> usize;
+    fn reset(&mut self) {}
```
BurntSushi: Can you please explain this? Preferably with documentation. (Bonus points for documenting `size`.)

BurntSushi:
@SeanRBurton This is a heroic clean up. I tried myself a few different times to get the shrinker impl'd in a macro so that we could make this work easily with multiple arities. Really nice job.

I think I have one doc nit and then a rebase and this should be good to go. Exciting!

SeanRBurton:
I merged this with master. It should now be good to go.

BurntSushi:
Yay! Awesome work. Thanks so much.

PR #102, opened 2015-10-22
Ramp size

SeanRBurton:
It makes sense to start the size off small and gradually ramp it up. Making this change uncovered some bugs relating to numeric overflow and gen_range bounds which I have now fixed.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -577,24 +577,16 @@ macro_rules! signed_shrinker {
 }
 
 macro_rules! signed_arbitrary {
-    ($(($ty:ident, $unsigned:ident)),*) => {
+    ($($ty:ident),*) => {
         $(
             impl Arbitrary for $ty {
                 fn arbitrary<G: Gen>(g: &mut G) -> $ty {
-                    let umax = ::std::$unsigned::MAX;
                     let s = g.size();
-                    let s = if s as $unsigned as usize == s {
-                        s as $unsigned
+                    if s as $ty as usize == s && s as $ty >= 0 {
+                        let s = s as $ty;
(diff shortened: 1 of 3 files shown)


BurntSushi:
This is a big change. Could you please describe in more detail what's going on?

SeanRBurton:
This includes some stuff from my other pull requests. I'm feeling a little confused about the etiquette around making multiple pull requests. Do you have a particular preference or a workflow that you use personally?

BurntSushi:
@SeanRBurton Here is my key concern: every pull request I accept means that I'm stating: yes, this is code I would like to maintain. In order to be OK with that, I have to understand what the code is doing and what problem it's solving. This PR contains quite a lot of changes, but a cryptic title and description that doesn't really help me understand the code, so it's hard for me to accept as is.

My personal preference is that you open issues discussing a big change like this first (similarly for auto-shrinking). Smallish PRs like your refactoring are totally fine to throw against the wall. I think the key goal of opening an issue first is: 1) making me understand the problem you're trying to solve and 2) convincing me that it's a problem worth solving. (I use "me" here because I'm the only maintainer of this project, but if there are more maintainers in the future, than "me" expands to include them as well.)

With all that said, it seems like this PR is meant to be fixing changes you made in #100? If so, that's a problem, because I didn't realize #100 needed fixing and I don't understand why it does now.

SeanRBurton:
OK, this should be much better. Sorry about the earlier confusion. For the record, all fixed bugs were pre-existing.

BurntSushi:
With the revert of #100, I think we need to go back to square 1 on this auto-shrinking stuff. It's my fault for not sticking to my guns and merging code that I didn't understand. If you want to help me work through it, I'd be happy to revisit.

PR #103, opened 2015-10-24
Add implementation for arbitrary arrays

hauleth:
Currently there is no shrink implementation for them as I have no idea how it should look like.

### DIFF
--- a/examples/reverse.rs
+++ b/examples/reverse.rs
@@ -11,8 +11,13 @@ fn reverse<T: Clone>(xs: &[T]) -> Vec<T> {
 }
 
 fn main() {
-    fn equality_after_applying_twice(xs: Vec<isize>) -> bool {
+    fn equality_after_applying_twice_vec(xs: Vec<isize>) -> bool {
         xs == reverse(&reverse(&xs))
     }
-    quickcheck(equality_after_applying_twice as fn(Vec<isize>) -> bool);
+    quickcheck(equality_after_applying_twice_vec as fn(Vec<isize>) -> bool);
+
+    fn equality_after_applying_twice_array(xs: [u8; 10]) -> bool {
+        &xs[..] == &reverse(&reverse(&xs))[..]
+    }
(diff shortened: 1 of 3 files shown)


examples/reverse.rs
```diff
@@ -11,8 +11,13 @@ fn reverse<T: Clone>(xs: &[T]) -> Vec<T> {
...
+    fn equality_after_applying_twice_vec(xs: Vec<isize>) -> bool {
         xs == reverse(&reverse(&xs))
     }
-    quickcheck(equality_after_applying_twice as fn(Vec<isize>) -> bool);
+    quickcheck(equality_after_applying_twice_vec as fn(Vec<isize>) -> bool);
+
+    fn equality_after_applying_twice_array(xs: [u8; 10]) -> bool {
+        &xs[..] == &reverse(&reverse(&xs))[..]
+    }
+    quickcheck(equality_after_applying_twice_array as fn([u8; 10]) -> bool);
```
BurntSushi: @hauleth Could you remove the arrays from this example? I'd like to keep this example short and sweet.

BurntSushi:
Great. Could you add at least one test for this?

BurntSushi:
@hauleth OK, so I'm kind of having second thoughts on this. Are you adding these impls to solve a particular problem? I'm kind of hesitant to add such things if they (1) aren't commonly used and (2) impact compile times. I also think that even if we do add them, they should come with a shrink implementation.

hauleth:
Sometimes I need buffers with fixed length when testing Octavo (passwords for symmetric ciphers and salts for KDFs). Of course i could try to implement shrink, but I don't know which behaviour of that would be the right one. 

--
Łukasz Jan Niemier

Dnia 27 sty 2016 o godz. 22:42 Andrew Gallant <notifications@github.com> napisał(a):

BurntSushi:
@hauleth You might consider new typing arrays and impl'ing `Arbitrary` on those?

As for shrinking, you obviously can't shrink the size of an array, but the *values* can be shrunk.

PR #104, opened 2015-10-27
Fix overflow bugs and add regression test.

SeanRBurton:
It's not safe to negate (e.g. call abs() on) the lowest possible value of a signed integral type.


BurntSushi:
Hmm, OK, if you rebase this I'd be happy to merge. Thanks!

PR #106, opened 2015-11-25
Use std::iter::empty inside empty_shrinker

bluss:
The `empty` iterator is a zero sized type, and this way the empty
shrinker (also the default shrinker) uses no allocation at all.

empty and once require Rust 1.2 or later.


BurntSushi:
QuickCheck pays no mind to allocation, but this still seems good to do anyway. Thanks!

bluss:
:smile: I guess, it came out of habit.

Issue #107, opened 2015-11-25
Hangs in shrinking / testcase failure


bluss:
Current git version seems to hang after a test failure, continuously spawning new threads.

bluss:
This is enough for me to reproduce a hang:

```
#[test]
fn qc_bug() {
    fn prop(a: Vec<u8>) -> bool {
        a.contains(&1)
    }
    quickcheck::quickcheck(prop as fn(_) -> bool);
}
```

bluss:
The bug is not in 49db7a7

cc @SeanRBurton due to recently merged PRs, maybe you know what's going on?

BurntSushi:
Yeah, I haven't had time to grok those PRs yet (which is why I haven't published a new release). If I can't find time today or tomorrow, I'll revert and revisit them later.

PR #108, opened 2015-11-25
Fix issue #107

SeanRBurton:
Sorry about this.

### DIFF
--- a/src/shrink.rs
+++ b/src/shrink.rs
@@ -303,10 +303,12 @@ impl Shrinker for StdShrinker {
             &mut StdShrinkerBody::Div8(ref mut shrinker) =>
                 apply_strategy!(shrinker, Div8),
             &mut StdShrinkerBody::Sub8(ref mut shrinker) if self.pass >= 4 => {
-                    return shrinker.use_shrinker(size, pool);
+                return shrinker.use_shrinker(size, pool);
             }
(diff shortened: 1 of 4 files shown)


BurntSushi:
@SeanRBurton Could you please explain the problem/fix? Thanks.

src/shrink.rs
```diff
@@ -303,10 +303,12 @@ impl Shrinker for StdShrinker {
             &mut StdShrinkerBody::Div8(ref mut shrinker) =>
```
bluss: I guess that's a fix for another time, but this would be in more usual style if all the `&mut` were dropped.

bluss:
you're on the ball, that was quick!

BurntSushi:
@SeanRBurton I'm not sure if you saw my previous comment, but could you please explain the fix in the commit message?

bluss:
@SeanRBurton Hi, it would be great if you could clarify this. I also think automatic shrinking sounds awesome (I've got some graphs that are way too complicated to write the shrinker for).

BurntSushi:
@SeanRBurton Unfortunately, I don't yet understand the auto shrinking code which means I can't maintain it. I unfortunately had to revert #100 as a consequence. I'd be happy to bring the changes back in, but I'll need your help understanding them first.

PR #110, opened 2015-12-02
Add support for std::time::{Duration, SystemTime}

### LINKED ISSUE #109: Add Arbitrary implementation for `std::time` (optionally)
Would be great to be able to generate these arbitrarily with `features = ["unstable"]`. It seems like this should probably `impl Arbitrary` for every reasonable owned type in the std library.

sgrif:
Arbitrary SystemTime can only be generated when built with `--features
unstable`. It doesn't seem like new arbitrary implementaions normally
get tested, so I've not added any tests for these. I also can't think of
any particularly important values for `shrink` here, as no durations or
times are particularly more likely to cause bugs than others.

We're moduloing the number of nanoseconds for a duration to always be
subsecond, so we don't get any isses when seconds is generated to
`u64::MAX_VAL`, or close to it.

Fixes #109.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ readme = "README.md"
 keywords = ["testing", "quickcheck", "property", "shrinking", "fuzz"]
 license = "Unlicense/MIT"
 
+[features]
+unstable = []
+
 [lib]
 name = "quickcheck"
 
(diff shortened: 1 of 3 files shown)


sgrif:
Looks like master is failing on nightly regardless of this branch (`rustc::plugin` is now `rustc_plugin`)

BurntSushi:
@sgrif Thanks! I think this looks OK, but I do think `shrink` should be defined. Shrinking is not only useful for finding the smallest input (so that it's easier to debug), but it's also useful as a heuristic for producing deterministic witnesses of failure. In an ideal world, re-running a failing quickcheck property will produce the same witness.

sgrif:
@BurntSushi I've added `shrink` definitions. Does that look correct to you?

BurntSushi:
Perfect! Thanks. :-)

PR #111, opened 2015-12-02
Fix the travis build on nightly

sgrif:
As of https://github.com/rust-lang/rust/pull/30043, `rustc::plugin` is
now `rustc_plugin`.


BurntSushi:
Awesome, thanks!

PR #113, opened 2015-12-06
Fix showing the message after panic from panic!() or assert!()

### LINKED ISSUE #91: Debug trait somehow getting lost (?) making debugging difficult
## Expected

Details about why my my data structure failed.

## Actual

> UNABLE TO SHOW RESULT OF PANIC.

## Details

(I am still learning Rust, so bear with me if my code is dumb or I miss something obvious.)

I am trying to define simple, stable-compatible quickcheck tests for a trait, that can be reused to test different data structures that implement it. In this case, I am trying to test a persistent heap.

[Data structure + generic test.](https://github.com/kazagistar/okasakish/blob/21a43f45d9cbc3633fef7510b8b21a607bdaf2f6/src/heap.rs)

Unfortunately, one of my two implementations is broken somewhere, but the only information I get out of quickcheck is `Error: "UNABLE TO SHOW RESULT OF PANIC."', <path>/quickcheck-0.2.21/src/tester.rs:113`.

[Link to broken build.](https://travis-ci.org/kazagistar/okasakish/jobs/76832591)

[Link to broken usage location](https://github.com/kazagistar/okasakish/blob/21a43f45d9cbc3633fef7510b8b21a607bdaf2f6/src/binomial_heap.rs#L145)

The data structure in question has `#[derive(Debug)]` and I have tested that `println("{:?}", s)` works for this instance of the data structure, so I don't know why I am not getting any other information about the panic, and I don't know what to do to make the situation better.

bluss:
Fix showing the message after panic from panic!() or assert!()

Typical panics produce either &str or String (depending if formatting is
used or not), so we use a simple type switch to handle these two after
a panic in a test case.

Fixes #91

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -414,9 +414,14 @@ fn safe<T, F>(fun: F) -> Result<T, String>
         where F: FnOnce() -> T, F: Send + 'static, T: Send + 'static {
     let t = ::std::thread::Builder::new().name("safe".into());
     t.spawn(fun).unwrap().join().map_err(|any_err| {
-        match any_err.downcast_ref::<&Debug>() {
-            Some(ref s) => format!("{:?}", s),
-            None => "UNABLE TO SHOW RESULT OF PANIC.".into(),
+        // Extract common types of panic payload:
+        // panic and assert produce &str or String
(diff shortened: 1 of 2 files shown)


BurntSushi:
Yay! Thanks. :-)

PR #125, opened 2016-02-28
Add Restriction to filter Arbitrary values

habnabit:
> Sometimes you want something arbitrary, but not too arbitrary. When you're encoding the something you want using the type signature of a #[quickcheck] test function, this is an easy way to skip values that you don't actually want without having to do a whole new Arbitrary implementation for the constraint.

Not complete yet since it lacks docs, but I'd like to know if this seems like a good approach at least.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -1,4 +1,4 @@
-use std::char;
+use std::{char, fmt};
 use std::collections::{
     BTreeMap,
     BTreeSet,
@@ -10,6 +10,7 @@ use std::collections::{
 };
 use std::hash::Hash;
 use std::iter::{empty, once};
+use std::marker::PhantomData;
 use std::ops::{Range, RangeFrom, RangeTo, RangeFull};
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


BurntSushi:
Hmm, could you describe a bit more about the problem you're trying to solve?

My first reaction to it is that it looks like a cute use of the type system, but that quickcheck is already too complex as it is. I'd prefer not expanding much beyond its current public API, other than perhaps more impls of standard types. (It does look like this could be prototyped in another crate.)

habnabit:
When writing tests for Aatch/ramp#60, I needed an integer value, but not a negative value. Since `#[quickcheck]` tests (as far as I can tell) are only able to request restrictions on their input parameters via the type system, this seemed like the only option.

BurntSushi:
@habnabit How is this exactly the only option? It seems like a newtype called `PositiveInt` or something similar would work OK? (It's generally what I do in such circumstances.)

Perhaps unrelated: you can discard test results and quickcheck will ignore that input as part of filling its test quota (up to some other larger limit). This probably isn't appropriate for your specific use case though, since negative integers will occur too frequently.

habnabit:
Sorry, what do you mean by "discard test results"?

`Restriction` was my solution to generalizing the boilerplate that's required when making the newtype, since otherwise you have to implement `Arbitrary` on every newtype. If all you're doing in `Arbitrary::arbitrary` is filtering values provided by your wrapped type, `Restriction`/`Restricted` provides that automatically with much less boilerplate for defining the predicate(s) to filter on.

BurntSushi:
When writing a property, if the return type is `TestResult`, then you can return a discarded test result: http://burntsushi.net/rustdoc/quickcheck/struct.TestResult.html#method.discard --- It's a way to tell quickcheck to ignore the given input without the particular test passing or failing.

I think I'd rather see this prototyped out of crate. If it turns out that a lot of people wind up finding it helpful, then I think I could be persuaded to bring it in. I otherwise want to be very very careful about expanding the current API.

PR #128, opened 2016-03-09
time2 has been stabilized

sgrif:
`duration_from_earlier` was renamed to `duration_since`, and the time2
feature is no longer needed. This can be removed from the unstable
feature gate once 1.8 is released, but these changes are required to
compile on beta and nightly.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -700,7 +700,7 @@ mod unstable_impls {
         }
 
         fn shrink(&self) -> Box<Iterator<Item=Self>> {
-            let duration = match self.duration_from_earlier(UNIX_EPOCH) {
+            let duration = match self.duration_since(UNIX_EPOCH) {
                 Ok(duration) => duration,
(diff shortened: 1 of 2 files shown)


sgrif:
@BurntSushi Any issues with this? This is failing Diesel's suite ATM, and I'd prefer to avoid pointing at a fork.

BurntSushi:
Thanks! Sorry about the delay.

> This can be removed from the unstable
feature gate once 1.8 is released

I'm not quite comfortable doing that, since that would break quickcheck on every Rust <1.8. I think the prevailing practice has been to maintain compatibility for at least a few stable releases. (This will get more interesting, I think, once distros start packaging Rust.)

> Unfortunately, quickcheck has been
unresponsive on updating to address this.

Personally, I think a response time within a few days is reasonable. If you need something quicker, then pointing at a fork might be necessary.

BurntSushi:
@sgrif This patch is now in `0.2.27` on crates.io. Thanks again!

Issue #127, opened 2016-03-09
Shrinking only appears to consider one "layer" of shrinking.


cstorey:
I'm used to clojure's `test.check`, which uses a recursive approach to shrinking, and I think it would make a nice addition, and is especially useful when generating a series of commands. For example, input to a state machine that models a replication instance.

For example, if we have a (hypothetical) sequence of operations modeled as booleans like: `[true, false, false, true, false, true]`, and we want to find the smallest sequence that contains more than 2 `true` instances, then we'd like to be able to reduce it down to a single `[true, true, true]` instance. Concretely:

```rust
    #[test]
    fn testy_test() {
        use quickcheck::Arbitrary;
        fn test_can_read_prepared_values_qc(vals: Vec<bool>) -> bool {
            vals.iter().filter(|&v| *v).count() < 3
        }
        quickcheck(test_can_read_prepared_values_qc as fn(vals: Vec<bool>) -> bool)
    }
```

At the moment, this results in (for example):
`        thread 'test::testy_test' panicked at '[quickcheck] TEST FAILED. Arguments: ([false, true, true, true, true, false, false, false, true, false, true, false, true, false, false, true, false, false, false, false, true, true, true, false, false, false, true, false, true, true, false, true, true, true, true, true, true, true, true, false, true, false, true, true, false, false, true, false, true, true, true, false, false, true, true, true, true, false, true, false, true, true, true, true, true, true, false, true, true, false, false, false, false, false, false, false, false])', /home/ceri/2016/03/09/quickcheck/src/tester.rs:116`

My feeling is that this shouldn't require too much surgery, should be confined to the `Testable` instances for functions.

cstorey:
As it turns out, the old code did this, by just recursing through `testable::shrink_failure`.

cstorey:
Fixed in #129

Issue #126, opened 2016-03-09
Regression in Shrinking


cstorey:
Using the following test case:
```rust
#[cfg(test)]
extern crate quickcheck;

#[cfg(test)]
mod test {
    use quickcheck::quickcheck;
    #[test]
    fn test_can_read_prepared_values() {
        fn test_can_read_prepared_values_qc(vals: Vec<Vec<u8>>) -> bool {
            false
        }
        quickcheck(test_can_read_prepared_values_qc as fn(vals: Vec<Vec<u8>>) -> bool)
    }
}
```

It seems to be that with version `0.2.24`, shrinking works okay, and we get:
```
---- replication_log::test::test_can_read_prepared_values stdout ----
        thread 'replication_log::test::test_can_read_prepared_values' panicked at '[quickcheck] TEST FAILED. Arguments: ([])', /home/ceri/.multirust/toolchains/stable/cargo/registry/src/github.com-88ac128001ac3a9a/quickcheck-0.2.24/src/tester.rs:113
```
When on `0.2.25`, I observe:
```
---- replication_log::test::test_can_read_prepared_values stdout ----
        thread 'replication_log::test::test_can_read_prepared_values' panicked at '[quickcheck] TEST FAILED. Arguments: ([[92, 51, 31, 18, 37, 45, 86, 89, 0, 27, 14, 52, 11, 4
1, 57, 65, 5, 65, 83, 87, 92, 36, 38, 18, 90, 85], ...])', /home/ceri/.multirust/toolchains/stable/cargo/registry/src/github.com-88ac128001ac3a9a/quickcheck-0.2.25/src/tester.rs:116
```

I plan to have a look myself, but It seems worth noting anyway.

BurntSushi:
I'll have a look at this soon, but if someone beats me to it, I might start looking here: https://github.com/BurntSushi/quickcheck/compare/0.2.24...0.2.25#diff-a7670f864b31d4c28b62524be9e1745bR274 --- The specific change there was defining a macro to make adding more `Testable` impls on functions with bigger arity easier. It looks like the code that handles failures is quite a bit different.

There was an overhaul to the `Vec` shrinker, but that appears to have been done before 0.2.24.

cstorey:
That's fair; `git bisect` lead me to commit `f017c01`, which is pretty much as you describe. I've been able to confirm too that the shrinker does still generate the same output on both versions, so that seems likely. 

[...]

PR #129, opened 2016-03-09
Port over the shrinking behavior from before f017c01.

cstorey:
Re: tickets #126 and #127.

This ensures that we try to recursively shrink a failing case, in order to find the smallest instance. Repeating myself, this means that the test case:

```rust
#[cfg(test)]
mod test {
    use quickcheck::quickcheck;
    #[test]
    fn testy_test() {
        use quickcheck::Arbitrary;
        fn thetest(vals: Vec<bool>) -> bool {
            vals.iter().filter(|&v| *v).count() < 2
        }
        quickcheck(thetest as fn(vals: Vec<bool>) -> bool)
    }
```

Will fail with `[quickcheck] TEST FAILED. Arguments: ([true, true])`, rather than something like `[quickcheck] TEST FAILED. Arguments: ([false, false, false, true, true, true, false, true, true, ...])`. This was mostly just adapting the logic used previously to the new macro.

I'd welcome any feedback on how you think it might be improved.

Thanks,

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -335,3 +335,18 @@ fn safe<T, F>(fun: F) -> Result<T, String>
 /// Convenient aliases.
 trait AShow : Arbitrary + Debug {}
 impl<A: Arbitrary + Debug> AShow for A {}
+
+#[cfg(test)]
+mod test {
+    use {QuickCheck, StdGen, quickcheck};
+    use super::TestResult;
+    #[test]
+    fn shrinking_regression_issue_126() {
+        fn thetest(vals: Vec<bool>) -> bool {
+            vals.iter().filter(|&v| *v).count() < 2
+        }
+        let failing_case = QuickCheck::new().quicktest(thetest as fn(vals: Vec<bool>) -> bool).unwrap_err();
+        let expected_argument = format!("{:?}", [true, true]);
+        assert_eq!(failing_case.arguments, vec![expected_argument]);
+    }
+}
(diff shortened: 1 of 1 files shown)


cstorey:
FWIW, the tests seem to be failing thanks to some oddities with[[how travis deploy the rust compiler](https://travis-ci.org/BurntSushi/quickcheck/jobs/114837133). They pass on both nightly and stable locally.

BurntSushi:
Thanks! Is there any chance we can add a regression test for this? I've found it pretty tricky in general to test this stuff, but one way might be to take your test and run it with [`quicktest`](http://burntsushi.net/rustdoc/quickcheck/struct.QuickCheck.html#method.quicktest) (which returns a `Result<usize, TestResult>`, unlike the typically more convenient `quickcheck`). You'd then have to inspect the `Vec<String>` arguments: http://burntsushi.net/rustdoc/src/quickcheck/tester.rs.html#130-134

If not, no worries and I'll get this merged today.

cstorey:
I'll more than likely have some time to do that tomorrow, then.

cstorey:
Okay, I've added the test; but I'll happily admit that it is basically a hack; hypothetically, I could parameterize the `TestResult` on the argument-tuple type, but that'd be a significant change for what seems like comparatively little value.

Thanks,

BurntSushi:
@cstorey Awesome, that looks like exactly what I had in mind! Thank you! :-)

BurntSushi:
This PR is in `0.2.26` on crates.io.

PR #132, opened 2016-05-10
Proper type annotation for modified prop function

badboy:


gaborigloi:
I think the `Vec<int>`s should also be corrected in the README to `Vec<isize>`.

badboy:
That's what you get for not testing the README :D 
Fixed it.

gaborigloi:
Yeah, this is why I really like the idea of https://github.com/brson/rust-skeptic, which can test the  README :).

BurntSushi:
Thanks!

PR #145, opened 2016-10-01
Add crate feature "use_logging" that's enabled by default

### LINKED ISSUE #142: Drop dependency on env_logger
env_logger pulls in regex, which is quite a lot to compile for a project where you just want smart tests.

I consider it normal to use quickcheck everywhere, just like I have benchmarks everywhere. If quickcheck is in the project, all of regex must be compiled in release mode as well to run benchmarks. For this reason I think the current deps are too large.

bluss:
This is to allow a faster compile of quickcheck. Logging brings a lot of dependencies that are more or less unused.

Fixes #142

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,11 +12,13 @@ license = "Unlicense/MIT"
 
 [features]
 unstable = []
+use_logging = ["log", "env_logger"]
+default = ["use_logging"]
 
 [lib]
 name = "quickcheck"
 
 [dependencies]
-env_logger = "0.3"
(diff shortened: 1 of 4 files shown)


bluss:
Feel free to decide a better name or a better way to do this.

I'm a bit sensitive to the size of deps. I use quickcheck a lot, I compile quickchecks in both debug and release mode (often release mode to have them finish in reasonable time). I switch nightly versions; it ends up being a lot of compiling of that excellent but redundant regex dependency.

PR #148, opened 2016-10-21
Allow QC_TESTS and QC_MAX_TESTS to set tests / max_tests

blt:
In practice I've wanted two kinds of quickcheck runs: fast and
comprehensive. To date the comprehensive testing has been done
by manually adjusting tests / max_tests up from the defaults of
100 / 10000. This involves a recompilation step.

If quickcheck were able to read environment variables to set its
tests / max_tests that recompilation step would not be needed and
a nightly CI task could be written to do a comprehensive QC run.

This commit introduces QC_TESTS and QC_MAX_TESTS to meet the
above. In the event that QC_MAX_TESTS < QC_TESTS then the
max_tests will be set to QC_TESTS.

Signed-off-by: Brian L. Troutwine blt@postmates.com

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -1,5 +1,7 @@
 use std::fmt::Debug;
 use std::panic;
+use std::env;
+use std::cmp;
 
 use rand;
 
@@ -13,6 +15,21 @@ pub struct QuickCheck<G> {
     gen: G,
 }
 
+fn qc_tests() -> usize {
+    match env::var("QC_TESTS") {
+        Ok(val) => val.parse().expect("QC_TESTS value could not converted to number"),
+        Err(_) => 100,
+    }
+}
+
+fn qc_max_tests() -> usize {
+    match env::var("QC_MAX_TESTS") {
+        Ok(val) => val.parse().expect("QC_MAX_TESTS value could not converted to number"),
+        Err(_) => 10000,
+    }
+}
+
+
 impl QuickCheck<StdGen<rand::ThreadRng>> {
     /// Creates a new QuickCheck value.
     ///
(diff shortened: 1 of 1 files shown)


src/tester.rs
```diff
@@ -13,6 +15,21 @@ pub struct QuickCheck<G> {
     gen: G,
 }
 
+fn qc_tests() -> usize {
+    match env::var("QC_TESTS") {
+        Ok(val) => val.parse().expect("QC_TESTS value could not converted to number"),
```
BurntSushi: If the parse fails, we *definitely* shouldn't panic here. I think we should just return the default value if you can't parse a number. (Alternatively, we could make `QuickCheck::new` return a result, but I think just returning the default on a parse failure is fine.)

Similarly for `qc_max_tests`.

BurntSushi:
This is a good idea. I have one nit and one question: do we want to use `QC` instead of the long form `QUICKCHECK`? I kind of feel like it should be the long form name.

blt:
Cool. I'll fix the panic. Initial motivation was that panicking would give the user a quick indication of a parse failure. Perhaps an error log instead? 

I can get behind long-form "QUICKCHECK".

BurntSushi:
For the most part, if a user sees a panic message, then I tend to consider that a bug.

You could add an error log using the `log` crate. Note though that you'll need to hide it behind the new `use_logging` feature.

blt:
Hi @BurntSushi, I've incorporated your feedback and included an additional environment variable for controlling `Gen` size. I didn't add in the logging facility. I'd be happy to do so; the current state of things is sufficient for my use-case.

BurntSushi:
Fine with me, thanks! :-)

PR #152, opened 2016-12-07
Added attr support to quickcheck! macro.

emallson:
I wanted to be able to `#[ignore]` some of my quickcheck tests defined with `quickcheck!` because they're expensive and I don't want to run them every time in dev. This (short) PR adds that ability.

I *don't* have tests for this, because I wasn't sure how to go about doing so. All existing tests pass, however.

    quickcheck! {
        #[ignore]
        fn prop(...) -> bool {
            ...
        }
    }

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,6 +46,7 @@ macro_rules! quickcheck {
     (@as_items $($i:item)*) => ($($i)*);
     {
         $(
+            $(#[$m:meta])*
             fn $fn_name:ident($($arg_name:ident : $arg_ty:ty),*) -> $ret:ty {
                 $($code:tt)*
             }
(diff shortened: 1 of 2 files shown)


BurntSushi:
Ah this sounds like a good change! I don't *think* it's a breaking change. cc @bluss

I think if you just add a test [like this one](https://github.com/BurntSushi/quickcheck/blob/master/src/tests.rs#L28) that uses the `#[ignore]` attribute, then that should be sufficient. I think that if you did that *without* this change, then tests would fail to compile, right?

emallson:
@BurntSushi I had forgotten about the `#[should_panic]` attribute. I used that with a test containing `assert!(false)`. As long as the attributes work, it will pass.

BurntSushi:
Ooo good idea! I'll wait a bit and see if @bluss wants to chime in but I think this otherwise good to go! Thanks so much!

bluss:
The macro changes look good, I think it's a good idea too!

BurntSushi:
Sweet off we go!

BurntSushi:
`0.4.1` is on crates.io.

PR #154, opened 2017-01-08
Implement `Arbitrary` for `Box`

bryal:
Lots of boilerplate is required when implementing `Arbitrary` for recursive types.
Having an impl of `Arbitrary` for boxes would be quite nice when working with such structures!

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -674,6 +674,15 @@ impl Arbitrary for Duration {
     }
 }
 
+impl<A: Arbitrary> Arbitrary for Box<A> {
+    fn arbitrary<G: Gen>(g: &mut G) -> Box<A> {
+        Box::new(A::arbitrary(g))
+    }
+
+    fn shrink(&self) -> Box<Iterator<Item=Box<A>>> {
+        Box::new((**self).shrink().map(Box::new))
+    }
+}
 
 #[cfg(feature = "unstable")]
 mod unstable_impls {


BurntSushi:
Ah yeah this is great! I'm surprised we've made it this far without it. :-)

BurntSushi:
OK, so I initially merged this and put out a `0.4.2` release of quickcheck. However, @bluss let me know that this is (unfortunately) a breaking change. It doesn't seem like it would be, but [`Box` is special](https://github.com/rust-lang/rfcs/blob/master/text/1023-rebalancing-coherence.md):

> A #[fundamental] type Foo is one where implementing a blanket impl over Foo is a breaking change. As described, & and &mut are fundamental. This attribute would be applied to Box, making Box behave the same as & and &mut with respect to coherence.

I completely forgot about `Box` being `#[fundamental]`.

This means we either need to revert this change or just make a new `0.5` release. I'm OK with a new `0.5` release.

BurntSushi:
From @bluss: Here's [an example](https://play.rust-lang.org/?code=impl%3CN%2C%20E%2C%20Ty%2C%20Ix%3E%20Arbitrary%20for%20Box%3CGraph%3CN%2C%20E%2C%20Ty%2C%20Ix%3E%3E%0A%20%20%20%20where%20N%3A%20Arbitrary%2C%0A%20%20%20%20%20%20%20%20%20%20E%3A%20Arbitrary%2C%0A%20%20%20%20%20%20%20%20%20%20Ty%3A%20EdgeType%20%2B%20Send%20%2B%20%27static%2C%0A%20%20%20%20%20%20%20%20%20%20Ix%3A%20IndexType%20%2B%20Send%2C%0A%7B%0A%20%20%20%20fn%20arbitrary%3CG%3A%20Gen%3E(g%3A%20%26mut%20G)%20-%3E%20Self%20%7B%0A%20%20%20%20%20%20%20%20Box%3A%3Anew(%3C_%3E%3A%3Aarbitrary(g))%0A%20%20%20%20%7D%0A%7D%0A%0A&version=stable&backtrace=0) of code that would work on `quickcheck 0.4.1` but would break with this patch applied.

BurntSushi:
I've yanked `quickcheck 0.4.2`.

bluss:
Yeah, I think that's for the best.

The rule summary is
- New blanket impls over Box, &, &mut are breaking changes.

The rules are what permit `impl RemoteTrait for Box<LocalType>` or `impl<'a> RemoteTrait for &'a LocalType` etc to compile, if the blanket impls don't exist in the remote crate.

bryal:
Interesting, hadn't thought of that! I think a `0.5` release sounds good.

PR #157, opened 2017-02-09
Handle non-terminating shrinkers in shrink_failure

JacekLach:
Rather than failing with a stack overflow, if we hit an arbitrary
depth, we report the most simplified form, and note that shrinking
does not seem to terminate.

This is a more generic alternative to #139, and should address common cases of #133.

Before this is ready to merge (assuming it's the approach you want to go with), we need to decide what a reasonable default recursion limit (maximum depth) is. Currently it's hardcoded to 1000 in an embarrassing fashion. I'm not sure what the best way of exposing it to configuration in `QuickCheck<G>` is; perhaps it should just be an additional argument to `.result()`, passed in by `quicktest<A>`?

Disclaimer: this is the first piece of Rust I've ever written, and was mostly done so that I have something to experiment with. So I might be missing something obvious, doing something really wrong, etc; you should probably review it more closely than something from an experienced rust dev. But hey, it compiled and passed tests!

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -301,25 +301,41 @@ impl<T: Testable,
             self_: fn($($name),*) -> T,
             a: ($($name,)*),
         ) -> Option<TestResult> {
-            for t in a.shrink() {
-                let ($($name,)*) = t.clone();
+            let mut values = Vec::new();
+            let mut best = None;
+            let mut depth = 1;
+
+            values.extend(a.shrink());
+            while let Some(next) = values.pop() {
+                let ($($name,)*) = next.clone();
                 let mut r_new = safe(move || {self_($($name),*)}).result(g);
                 if r_new.is_failure() {
                     {
-                        let ($(ref $name,)*) : ($($name,)*) = t;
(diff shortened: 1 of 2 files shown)


src/tester.rs
```diff
@@ -301,25 +301,39 @@ impl<T: Testable,
             self_: fn($($name),*) -> T,
             a: ($($name,)*),
         ) -> Option<TestResult> {
-            for t in a.shrink() {
-                let ($($name,)*) = t.clone();
+            let mut values = a.shrink().collect::<Vec<_>>();
```
bluss: Is it feasible to collect all shrinks? I'm thinking it should be possible to feed shrink a very very large list.

JacekLach: Good catch, and I can resolve that easily - the realisation of the iterator was only necessary back when I thought this was doing a full depth search (with backtracking). Since we don't do that, we can just use an iterator in the while loop, and reassign it when we descend.

BurntSushi:
Closing due to inactivity. Thanks for attempting this though!

PR #164, opened 2017-03-12
unbreak attribute for functions with pattern parametres

strake:
e.g.

```
#[quickcheck] fn test(mut xv: Vec<i32>) {}
```

fails with E0561

### DIFF
--- a/quickcheck_macros/src/lib.rs
+++ b/quickcheck_macros/src/lib.rs
@@ -11,7 +11,7 @@ extern crate syntax;
 extern crate rustc_plugin;
 
 use syntax::ast;
-use syntax::ast::{Ident, ItemKind, StmtKind, Stmt, TyKind};
+use syntax::ast::{Ident, ItemKind, PatKind, StmtKind, Stmt, TyKind};
 use syntax::codemap;
 use syntax::ext::base::{ExtCtxt, MultiModifier, Annotatable};
 use syntax::ext::build::AstBuilder;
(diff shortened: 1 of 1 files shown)


strake:
I believe the broken build is due to broken nightly rather than this patch.

BurntSushi:
_kicks CI_

strake:
Will you have time to review this soon? It's breaking my CI builds. (It's a regression, by the way.)

BurntSushi:
@strake `quickcheck_macros 0.4.2` is now on crates.io with this patch. Thanks!

FYI, if you're concerned about your CI building, then you really shouldn't be using `quickcheck_macros`. It only works with a nightly compiler and is therefore very unstable. I can't promise to you that I will always be able to keep up with the breakage. You may find that the [`quickcheck!`](https://docs.rs/quickcheck/0.4.2/quickcheck/macro.quickcheck.html) macro, which works on stable, is good enough.

bluss:
@strake You could pick up and PR this improvement to the macro. Then it will support your `mut v` parameter. BurntSushi/quickcheck/issues/144

strake:
@BurntSushi: Nah, i pin the nightly version so it's fine at least until i want to upgrade.

@bluss: Yeah, i looked at that too, but alas ':' can't follow a pattern in a macro.

bluss:
It works fine for me though

bluss:
The macro that is linked in that issue https://github.com/bluss/rust-itertools/blob/a048d32fb05d12f75ba36ae8f285204a08607255/tests/quick.rs#L171-L209  (and the first use of it uses `mut` in the pattern).

PR #166, opened 2017-03-29
Add Arbitrary for non-default Hashers in HashMap/HashSet

emallson:
The motivation for this is that I want to use a non-standard hasher (in my case, [fnv](https://doc.servo.org/fnv/)) instead of the default `SipHasher`. However, the current implementation has trait bounds that implicitly restrict the third parameter of `HashMap` (or second of `HashSet`) to the [default](https://doc.rust-lang.org/std/collections/hash_map/struct.RandomState.html).

It turned out to be very simple to address this: parameterizing the `impl`'s with the hasher type parameter `S` works with both the short-hand generally used (`HashMap<K, V>` or `HashSet<K>`) and with types that have the third specified (e.g. `HashMap<K, V, BuildHasherDefault<FnvHasher>>`).

The added tests are purely compilation tests: if they compile, they pass. One set checks that the normal `HashMap<K, V>` types still work, and the other tests that adding the parameter for the `Hasher` works. I didn't want to add a new dev-dependency, so it simply uses the built-in `SipHasher`. While this is basically equivalent to the default, it does fail without modifying the implementation of `Arbitrary` for these types because `RandomState` (the default) is not a type alias for `BuildHasherDefault<SipHasher>`.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -8,7 +8,7 @@ use std::collections::{
     LinkedList,
     VecDeque,
 };
-use std::hash::Hash;
+use std::hash::{BuildHasher, Hash};
 use std::iter::{empty, once};
 use std::ops::{Range, RangeFrom, RangeTo, RangeFull};
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


BurntSushi:
@emallson Ah this is awesome, thanks so much! I think those `Arbitrary` impls were written before those types ever got a third type parameter. :-)

PR #168, opened 2017-04-17
Arbitrary for PathBuf and OsString

droundy:
Hello, I just implemented hopefully a reasonable Arbitrary for PathBuf (and incidentally OsString).  The PathBuf arbitrary tries to get somewhat interesting and/or realistic paths, but doesn't try to hard (e.g. it will *extremely** rarely pick a path like foo/../bar.  But it does get realistic prefixes (hopefully), and shouldn't do anything too insane.

I also added *~ to gitignore.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -1,3 +1,4 @@
+use std;
 use std::char;
 use std::collections::{
     BTreeMap,
@@ -12,6 +13,9 @@ use std::hash::{BuildHasher, Hash};
 use std::iter::{empty, once};
 use std::ops::{Range, RangeFrom, RangeTo, RangeFull};
 use std::time::Duration;
+
+use std::ffi::OsString;
+use std::path::PathBuf;
 
 use rand::Rng;
 
@@ -388,6 +392,42 @@ impl<T: Arbitrary> Arbitrary for VecDeque<T> {
     }
 }
 
+impl Arbitrary for PathBuf {
+    fn arbitrary<G: Gen>(g: &mut G) -> PathBuf {
+        // use some real directories as guesses, so we may end up with
+        // actual working directories in case that is relevant.
+        let here = std::env::current_dir().unwrap_or(PathBuf::from("/test/directory"));
+        let temp = std::env::temp_dir();
+        let home = std::env::home_dir().unwrap_or(PathBuf::from("/home/user"));
+        let mut p = g.choose(&[here,
+                               temp,
+                               home,
+                               PathBuf::from("."),
+                               PathBuf::from(".."),
+                               PathBuf::from("../../.."),
+                               PathBuf::new()]).unwrap().clone();
+        for sub in Vec::<OsString>::arbitrary(g).iter() {
+            p.push(sub);
+        }
+        p
+    }
+
+    // I'm not sure how to shrink this.  Help appreciated!
+    // fn shrink(&self) -> Box<Iterator<Item=PathBuf>> {
+    // }
+}
+
+impl Arbitrary for OsString {
+    fn arbitrary<G: Gen>(g: &mut G) -> OsString {
+        OsString::from(String::arbitrary(g))
+    }
+
+    fn shrink(&self) -> Box<Iterator<Item=OsString>> {
+        let mystring: String = self.clone().into_string().unwrap();
+        Box::new(mystring.shrink().map(|s| OsString::from(s)))
+    }
+}
+
 impl Arbitrary for String {
     fn arbitrary<G: Gen>(g: &mut G) -> String {
         let size = { let s = g.size(); g.gen_range(0, s) };
@@ -476,6 +516,7 @@ impl Arbitrary for char {
         Box::new((*self as u32).shrink().filter_map(char::from_u32))
     }
 }
+
 
 macro_rules! unsigned_shrinker {
     ($ty:ty) => {
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -3,6 +3,7 @@ use rand;
 use super::{QuickCheck, StdGen, TestResult, quickcheck};
 use std::collections::{HashSet, HashMap};
 use std::hash::{BuildHasherDefault, SipHasher};
+use std::path::PathBuf;
 
 #[test]
 fn prop_oob() {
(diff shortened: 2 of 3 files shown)


sdleffler:
I could really use this. Perhaps as a start to shrinking `PathBuf`, it could be `.pop()`'d and the `.pop()`'d path could be added to the shrinker if `.pop()` returned true?

src/arbitrary.rs
```diff
@@ -13,6 +14,9 @@ use std::iter::{empty, once};
 use std::ops::{Range, RangeFrom, RangeTo, RangeFull};
 use std::time::Duration;
 
+use std::ffi::OsString;
+use std::path::PathBuf;
```
BurntSushi: These should be grouped with the other `std` imports.

src/arbitrary.rs
```diff
@@ -388,6 +392,63 @@ impl<T: Arbitrary> Arbitrary for VecDeque<T> {
...
+        let mut p = g.choose(&[here,
+                               temp,
+                               home,
+                               PathBuf::from("."),
+                               PathBuf::from(".."),
+                               PathBuf::from("../../.."),
+                               PathBuf::new()]).unwrap().clone();
+        for sub in Vec::<OsString>::arbitrary(g).iter() {
+            p.push(sub);
+        }
```
BurntSushi: Seems like this could be replaced with `p.extend(Vec::<OsString>::arbitrary(g).iter())`.

src/arbitrary.rs
```diff
@@ -388,6 +392,63 @@ impl<T: Arbitrary> Arbitrary for VecDeque<T> {
...
+        let here = std::env::current_dir().unwrap_or(PathBuf::from("/test/directory"));
+        let temp = std::env::temp_dir();
+        let home = std::env::home_dir().unwrap_or(PathBuf::from("/home/user"));
+        let mut p = g.choose(&[here,
+                               temp,
+                               home,
+                               PathBuf::from("."),
+                               PathBuf::from(".."),
+                               PathBuf::from("../../.."),
+                               PathBuf::new()]).unwrap().clone();
```
BurntSushi: I'd prefer this style please:

```rust
let choices = [
    here,
    temp,
    ...
];
let mut p = g.choose(&choices).unwrap().clone();
```

src/arbitrary.rs
```diff
@@ -388,6 +392,63 @@ impl<T: Arbitrary> Arbitrary for VecDeque<T> {
     }
 }
 
+impl Arbitrary for PathBuf {
+    fn arbitrary<G: Gen>(g: &mut G) -> PathBuf {
+        // use some real directories as guesses, so we may end up with
+        // actual working directories in case that is relevant.
+        let here = std::env::current_dir().unwrap_or(PathBuf::from("/test/directory"));
+        let temp = std::env::temp_dir();
+        let home = std::env::home_dir().unwrap_or(PathBuf::from("/home/user"));
```
BurntSushi: This is clever. It feels a little weird to me, I admit, but I can't think of any concrete objections to this...

src/arbitrary.rs
```diff
@@ -388,6 +392,63 @@ impl<T: Arbitrary> Arbitrary for VecDeque<T> {
...
+                               PathBuf::from("../../.."),
+                               PathBuf::new()]).unwrap().clone();
+        for sub in Vec::<OsString>::arbitrary(g).iter() {
+            p.push(sub);
+        }
+        p
+    }
+
+    fn shrink(&self) -> Box<Iterator<Item=PathBuf>> {
+        let mut shrunk = Vec::new();
```
BurntSushi: Use `vec![]` please.

src/arbitrary.rs
```diff
@@ -388,6 +392,63 @@ impl<T: Arbitrary> Arbitrary for VecDeque<T> {
...
+
+        // Iterating over a Path performs a small amount of normalization.
+        let normalized = self.iter().collect::<PathBuf>();
+        if normalized.as_os_str() != self.as_os_str() {
+            shrunk.push(normalized);
+        }
+
+        // Add the canonicalized variant only if canonicalizing the path actually does something,
+        // making it (hopefully) smaller. Also, ignore canonicalization if canonicalization errors.
+        if let Ok(canonicalized) = self.canonicalize() {
```
BurntSushi: If `popped.pop()` returns `false`, what is the behavior of `canonicalize`?

src/arbitrary.rs
```diff
@@ -991,4 +1054,9 @@ mod test {
         ordered_eq(..3, vec![..0, ..2]);
         ordered_eq(.., vec![]);
     }
+
+    #[test]
+    fn pathbuf() {
+        ordered_eq(PathBuf::from("/home/foo//.././bar"), vec![PathBuf::from("/home/foo//.."), PathBuf::from("/home/foo/../bar")]);
```
BurntSushi: Please wrap code to 79 columns inclusive. Thanks.

src/arbitrary.rs
```diff
@@ -991,4 +1054,9 @@ mod test {
...
+    }
```
BurntSushi: Could we add an explicit test for `OsString` as well?

src/tests.rs
```diff
@@ -220,6 +221,13 @@ fn regression_issue_107_hang() {
 }
 
 quickcheck! {
+    /// The following is a very simplistic test, which only verifies
+    /// that our PathBuf::arbitraryl does not panic.  Still, that's
```
BurntSushi: `s/arbitraryl/arbitrary/`

src/arbitrary.rs
sdleffler: Frankly, I have no idea. I think it could vary. If popping takes the path down to `/` or some other root, `canonicalize` should return the same path; if popping removes the path entirely (no root) who knows what canonicalize would do - probably error. So, `canonicalize` (I think) has multiple possible behaviors in the case of `!popped.pop()`.

BurntSushi: Hmm OK, I think my primary concern is whether this bottoms out or not. e.g., It `canonicalize` keeps returning different paths, then it's possible that the shrinker never terminates.

BurntSushi: (But that seems like a quite surprising behavior, so I think we're fine as is.)

sdleffler: Hmm. I think it would be impossible to get stuck in a loop between `pop` and `canonicalize`. In order for `canonicalize` to normalize back to a different path, `pop` would need to pop back to a path which, when canonicalized, contains the original canonicalized path as a parent. `canonicalize` will result in wildly different paths when symbolic links are involved, and *should* remove things like `..` [citation needed]. So, I don't see a situation without a crazy looping hard link that will cause `canonicalize` and `pop` to cycle.

BurntSushi:
I addressed my feedback, squashed this down to one commit and merged it here: https://github.com/BurntSushi/quickcheck/commit/af95aabd9af57b3de6aed2067118cb1df9fc941b

PR #174, opened 2017-05-18
Adds Arbitrary implementation for Arc<Arbitrary>

remexre:
As copy-pasted from the Box implementation.


BurntSushi:
Nice! Thank you!

Issue #146, opened 2016-10-01
Support for 9, 10, 11, 12 tuples accounts for 35% of the compile time of the crate


bluss:
Support for 9, 10, 11, 12 tuples accounts for half of the compile time of the crate.

Diminishing returns. I would kill these off.

bluss:
- rustc 1.14.0-nightly (289f3a4ca 2016-09-29)
- quickcheck 0.3.1 6c7a4a85402e09301427700e9eed38c73a8f6cf9

code diff:

```diff
diff --git a/src/arbitrary.rs b/src/arbitrary.rs
index 422c07d..8fb836c 100644
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -192,14 +192,6 @@ impl_arb_for_tuple!((a, A), (b, B), (c, C), (d, D), (e, E), (f, F),
                     (g, G));
 impl_arb_for_tuple!((a, A), (b, B), (c, C), (d, D), (e, E), (f, F),
                     (g, G), (h, H));
-impl_arb_for_tuple!((a, A), (b, B), (c, C), (d, D), (e, E), (f, F),
-                    (g, G), (h, H), (i, I));
-impl_arb_for_tuple!((a, A), (b, B), (c, C), (d, D), (e, E), (f, F),
-                    (g, G), (h, H), (i, I), (j, J));
-impl_arb_for_tuple!((a, A), (b, B), (c, C), (d, D), (e, E), (f, F),
-                    (g, G), (h, H), (i, I), (j, J), (k, K));
-impl_arb_for_tuple!((a, A), (b, B), (c, C), (d, D), (e, E), (f, F),
-                    (g, G), (h, H), (i, I), (j, J), (k, K), (l, L));
 
 impl<A: Arbitrary> Arbitrary for Vec<A> {
     fn arbitrary<G: Gen>(g: &mut G) -> Vec<A> {
diff --git a/src/tester.rs b/src/tester.rs
index 8fed5b9..d6fe5ad 100644
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -311,10 +311,6 @@ testable_fn!(A, B, C, D, E);
 testable_fn!(A, B, C, D, E, F);
 testable_fn!(A, B, C, D, E, F, G);
 testable_fn!(A, B, C, D, E, F, G, H);
-testable_fn!(A, B, C, D, E, F, G, H, I);
-testable_fn!(A, B, C, D, E, F, G, H, I, J);
[...]

PR #175, opened 2017-05-29
#[derive(Arbitrary)]

remexre:
### Summary:
Implements a stable-friendly `#[derive(Arbitrary)]`, with support for generation, shrinking, and constraints. Example:

```rust
#[macro_use]
extern crate quickcheck;
#[macro_use]
extern crate quickcheck_derive;

#[derive(Arbitrary, Clone, Debug)]
#[arbitrary(constraint = "this.alpha == this.bravo.is_positive()")]
struct TestStruct {
    alpha: bool,
    bravo: isize,
}

quickcheck! {
    fn struct_constraint(t: TestStruct) -> bool {
        t.alpha == t.bravo.is_positive()
    }
}
```

### Known Issues:
 - More tests would probably be a good thing; the existing tests essentially just ensure the derive outputs code that compiles.
 - Struct shrinking is implementing by transforming a struct into a tuple, shrinking the tuple, and mapping it back to the struct. 9-tuples and larger cannot be shrunk (see #146), so structs with more than 8 members cannot either.
 - Constraints have to use `this` instead of `self`. Constraints get turned into a function defined inside the `arbitrary` method, where no `self` exists. They therefore cannot use `self` (at least, to my knowledge).

### DIFF
--- a/quickcheck_derive/src/lib.rs
+++ b/quickcheck_derive/src/lib.rs
@@ -1,0 +1,53 @@
+#![crate_type = "proc-macro"]
+#![recursion_limit = "128"]
+
+extern crate proc_macro;
+#[macro_use]
+extern crate quote;
+extern crate syn;
+
+mod attrs;
+mod structural;
+
+use attrs::*;
+use proc_macro::TokenStream;
+use structural::*;
+use syn::{Body, parse_derive_input};
+
+#[proc_macro_derive(Arbitrary, attributes(arbitrary))]
+pub fn derive(input: TokenStream) -> TokenStream {
+    let item = parse_derive_input(&input.to_string())
+        .expect("Couldn't parse item");
+    let (arbitrary, shrink) = match item.body {
+        Body::Struct(ref variant) => derive_struct(&item, &variant),
+        Body::Enum(ref variants) => derive_enum(&item, &variants),
+    };
+    let valid = process_attrs(&item);
+
+    let name = &item.ident;
+    let (impl_generics, ty_generics, where_clause) = item.generics.split_for_impl();
+    let ast = quote! {
+        impl ::quickcheck::Arbitrary for #impl_generics #name #ty_generics #where_clause {
+            #[allow(unused_mut, unused_variables)]
+            fn arbitrary<G: ::quickcheck::Gen>(_g: &mut G) -> Self {
+                // TODO Find a way to use "self" instead of "this".
+                let valid = |this: &Self| { #valid };
+                let mut gen = move || { #arbitrary };
+
+                loop {
+                    let out = gen();
+                    if valid(&out) {
+                        return out;
+                    }
+                }
+            }
+
+            fn shrink(&self) -> Box<Iterator<Item=Self>> {
+                #shrink
+            }
+        }
+    };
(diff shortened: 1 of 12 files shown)


quickcheck_derive/src/structural.rs
```diff
@@ -0,0 +1,170 @@
...
+        },
+        VariantData::Tuple(ref fields) => {
+            let field_names = (0..fields.len()).map(Ident::new).map(|i| quote!(self.#i));
+            let alpha_names = &alpha_names(fields.len());
+
+            quote! {
+                // TODO This isn't a *great* way to do this until we get
+                // generics over tuples, to be able to implement shrinking
+                // for tuples of all sizes.
+                Box::new((#(#field_names),*).shrink().map(|(#(#alpha_names),*)| #name(#(#alpha_names),*)))
```
Marwes: This looks really useful! Attempted to use it in a project of mine and discovered a bug with newtypes. `(#(#alpha_names),*)` needs to get the same treatement as `#tuple_pattern` above or it fails for `struct T(i32);`.

Marwes: This looks really useful! Attempted to use it in a project of mine and discovered a bug with newtypes. `(#(#alpha_names),*)` needs to get the same treatment as `#tuple_pattern` above or it fails for `struct T(i32);`.

(Accidently deleted this comment).

quickcheck_derive/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+    };
+    let valid = process_attrs(&item);
+
+    let name = &item.ident;
+    let (impl_generics, ty_generics, where_clause) = item.generics.split_for_impl();
+    let ast = quote! {
+        impl ::quickcheck::Arbitrary for #impl_generics #name #ty_generics #where_clause {
+            #[allow(unused_mut, unused_variables)]
+            fn arbitrary<G: ::quickcheck::Gen>(_g: &mut G) -> Self {
+                // TODO Find a way to use "self" instead of "this".
```
Marwes: You could probably introduce a trait which only exists in this method which has a `valid` method.

quickcheck_derive/src/structural.rs
remexre: Reproduced; I'll fix this and add a test for it tonight.

quickcheck_derive/src/lib.rs
remexre: Huh, didn't know a function-scoped trait was legal. I'll do that tonight, thanks!

Marwes:
> Struct shrinking is implementing by transforming a struct into a tuple, shrinking the tuple, and mapping it back to the struct. 9-tuples and larger cannot be shrunk (see #146), so structs with more than 8 members cannot either.

This could be fixed by nesting tuples ie, for a three field struct you could generate an arbitrary for `(A, (B, C))`

remexre:
@marcelbuesing It's fallen off my plate more or less completely; if someone else wants to take it across the finish line, that'll probably happen before I have time to come back to it.

BurntSushi:
https://github.com/BurntSushi/quickcheck/pull/222#issuecomment-573267560

PR #177, opened 2017-06-07
Add min_tests_passed

### LINKED ISSUE #37: Provide some way to fail the unit test if all the quickcheck tests were discarded
Sometimes I write too many "discard conditions" (along with too many arguments), or sometimes I get those conditions wrong, and quickcheck ends up discarding _all_ the `10k` tests, and flagging the unit test as _passed_. IMO, either case is a bug, because I'm not actually testing anything, and I should fix the discard conditions.

Right now, the only way to detect this issue is to use `RUST_LOG=quickcheck` and look out for "Passed 0 Quickcheck tests" messages.

I would like to see some way to fail the unit test if too few quickcheck tests passed. Preferably, using a environment variable, say `QC_PASSED_TESTS_THRESHOLD=25 cargo test` will fail the unit test if less than 25 quickcheck tests passed, if the variable is unset, the threshold will be zero, which is the current behavior.

Thoughts @BurntSushi ?

Jim-Holmstroem:
Fixes #37

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -12,6 +12,7 @@ use {Arbitrary, Gen, StdGen};
 pub struct QuickCheck<G> {
     tests: usize,
     max_tests: usize,
+    min_tests_passed: usize,
     gen: G,
 }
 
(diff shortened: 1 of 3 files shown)


src/tester.rs
```diff
@@ -39,6 +40,14 @@ fn qc_gen_size() -> usize {
     }
 }
 
+fn qc_min_tests_passed() -> usize {
+    let default = 0;
```
Jim-Holmstroem: Default value set to `0` to not break current behaviour.

BurntSushi:
@Jim-Holmstroem Could you please add `Fixes #37` to the commit message? Thanks. Otherwise, this looks good to me. Thanks!

BurntSushi:
I resolved conflicts and merged this in https://github.com/BurntSushi/quickcheck/commit/531a25a9f14bded6bf669858c77955c548bf6cef

Issue #27, opened 2014-09-21
Arbitrary f32 and f64's should include special values


danburkert:
Currently `INFINITY`, `NEG_INFINITY`, `NAN` and possibly `-0.0` are never generated by quickcheck. These values are valid floats, and should be tested for.

huonw:
Including [subnormals](https://en.wikipedia.org/wiki/Denormal_number) might be good too.

bekh6ex:
Related problem: `f64` and `f32` values are always in range between -100 and 100
```rust
quickcheck! {
    fn float_range(n: f64) -> bool {
        n < 100.0 && n > -100.0
    }
}
```
This test never fails.

bekh6ex:
Regarding Haskell's quickcheck and floats here is related discussion:
https://github.com/nick8325/quickcheck/issues/98

xfix:
> However for Float, the NaN values are problematic since they violate the typeclass laws.

Doesn't sound like an issue for Rust, just saying, due to `PartialEq` and `Eq` being separate.

PR #183, opened 2017-09-26
Generate special values for f32 and f64

bekh6ex:
Issue #27

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -39,6 +39,16 @@ fn qc_gen_size() -> usize {
     }
 }
 
+fn qc_gen_floats_finite_only() -> bool {
+    let default = false;
+    match env::var("QUICKCHECK_GENERATOR_FLOATS_FINITE_ONLY") {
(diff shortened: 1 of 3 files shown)


src/tester.rs
```diff
@@ -385,4 +397,57 @@ mod test {
...
+    fn when_floats_finite_only_flag_set_to_false_generates_non_finite_f32() {
+        let floats_finite_only = false;
+
+        let mut gen = StdGen::new(rand::thread_rng(), 100, floats_finite_only);
+        let mut non_finite_was_found = false;
+
+        for _ in 0..10000 {
+            let f = f32::arbitrary(&mut gen);
+            if !f.is_finite() {
+                non_finite_was_found = true;
```
bekh6ex: This test can theoretically create a false negative, but probability of that is (as far as I've managed to calculate) 1 of 500.
Maybe number of iterations should be increased, should it?

BurntSushi:
Closing due to inactivity.

PR #191, opened 2017-12-08
readme: add link to proptest crate

BurntSushi:
The proptest crate is well documented and provides a fresh take on
property based testing inspired by the Hypothesis framework for Python.
In particular, it appears to improve the shrinking story that can be
inconvenient to work with in QuickCheck.

See also: https://github.com/AltSysrq/proptest/issues/15

cc @AltSysrq --- Does the wording look OK to you? Any suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,19 @@ Crate features:
   `RUST_LOG`.
 
 
+### Alternative Rust crates for property testing
+
+The [`proptest`](https://docs.rs/proptest) crate is inspired by the
+[Hypothesis](http://hypothesis.works) framework for Python.
+You can read a comparison between `proptest` and `quickcheck`
+[here](https://docs.rs/proptest/*/proptest/#differences-between-quickcheck-and-proptest)
+and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,6 +130,19 @@ Crate features:
   `RUST_LOG`.
 
 
+### Alternative Rust crates for property testing
+
+The [`proptest`](https://docs.rs/proptest) crate is inspired by the
+[Hypothesis](http://hypothesis.works) framework for Python.
+You can read a comparison between `proptest` and `quickcheck`
+[here](https://docs.rs/proptest/0.3.1/proptest/#differences-between-quickcheck-and-proptest)
```
AltSysrq: Maybe this should use `*` for the version so it always goes to the latest docs? I just released 0.3.2, so if not, it should be updated to point to that version.

BurntSushi: Neat! I didn't know `*` was a thing. Fixed.

PR #192, opened 2017-12-08
Fix broken link

ethanpailes:
I noticed this link was broken while reading through the docs.

I'm not completely happy with the doc link I added because it is pinned to a specific version rather than the latest version, which I think would make more sense for a reference from the head of master. Do you happen to know if there is a know good way to deal with this problem? If not it might be worth asking the docs.rs maintainers to add the ability to name the latest version with a URI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ quickcheck(prop as fn(Vec<isize>) -> TestResult);
 So now our property returns a `TestResult`, which allows us to encode a bit
 more information. There are a few more
 [convenience functions defined for the `TestResult`
-type](http://burntsushi.net/rustdoc/quickcheck/struct.TestResult.html).
+type](http://docs.rs/quickcheck/0.5.0/quickcheck/struct.TestResult.html).
 For example, we can't just return a `bool`, so we convert a `bool` value to a
 `TestResult`.
 


ethanpailes:
Actually, I just noticed that you have burntsushi.net configured to redirect to the latest, which is a clever workaround. I suppose a more principled fix might be to make the struct link on burntsushi.net redirect to the latest version. If you want to go that route, this PR is not required.

BurntSushi:
Yeah I definitely want to migrate everything over to docs.rs. I setup the redirect in the likely event that I didn't catch everything. :-) Thanks!

ethanpailes:
I actually think the redirect is a really nice solution to the fact that docs.rs does not seem to have a way to talk about the latest version. When you cut a new point release you just have to update your redirect without having to add changes to the README. Of course it would be better if docs.rs had the ability.

ethanpailes:
Wait, it turns out I just forgot to read the manual. There is a way. I can go through and update links if you want.

BurntSushi:
@ethanpailes Technically, the links in the README should be versioned. That way, you always get the right docs no matter which version of the README you're reading. The "problem" is remembering to update all of the links when the version gets bumped. :)

ethanpailes:
I hadn't thought of the rooting-through-git-history use case. Now that you mention it, the current situation seems like the best approach. I guess release tooling would solve the issue? That seems like way too much overkill.

PR #198, opened 2017-12-25
Fix broken build of quickcheck_macros

doomrobo:
Build was broken due to updated structure of `ast::BareFnTy`. All I did was rename the `lifetimes` member to `generic_params`, which remains an empty vector in the code generation.

```
   Compiling quickcheck_macros v0.5.0 (file:///private/tmp/quickcheck/quickcheck_macros)
error[E0560]: struct `syntax::ast::BareFnTy` has no field named `lifetimes`
  --> src/lib.rs:61:17
   |
61 |                 lifetimes: vec![],
   |                 ^^^^^^^^^^ `syntax::ast::BareFnTy` does not have this field
   |
   = note: available fields are: `unsafety`, `abi`, `generic_params`, `decl`
error: aborting due to previous error
error: Could not compile `quickcheck_macros`.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quickcheck"
-version = "0.5.0"  #:version
+version = "0.5.1"  #:version
 authors = ["Andrew Gallant <jamslam@gmail.com>"]
 description = "Automatic property based testing with shrinking."
 documentation = "http://burntsushi.net/rustdoc/quickcheck/"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quickcheck"
-version = "0.5.1"  #:version
+version = "0.5.0"  #:version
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "quickcheck"
-version = "0.5.0"  #:version
+version = "0.5.1"  #:version
```
BurntSushi: Please don't version bump. I do this as part of releases.

doomrobo: Reverted. I only bumped because anyone on 0.5.0 will still have a broken build

BurntSushi:
Thanks!

BurntSushi:
`quickcheck 0.5.1` is on crates.io with this change. Thanks again!

BurntSushi:
I goofed. `rand` is a public dep, so I actually needed to do a major version bump. I yanked `0.5.1` and released `0.6.0`.

PR #200, opened 2018-01-23
Allow creation of a QuickCheck instance with any Gen

UnlawfulMonad:
Title really says it all. Adds a `with_gen(generator: G)` method to `QuickCheck<G>`. Allows the use of generators other than the thread local one.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -59,17 +59,9 @@ impl QuickCheck<StdGen<rand::ThreadRng>> {
     /// the max number of overall tests is set to `10000` and the generator
     /// is set to a `StdGen` with a default size of `100`.
     pub fn new() -> QuickCheck<StdGen<rand::ThreadRng>> {
-        let tests = qc_tests();
-        let max_tests = cmp::max(tests, qc_max_tests());
         let gen_size = qc_gen_size();
-        let min_tests_passed = qc_min_tests_passed();
 
-        QuickCheck {
-            tests: tests,
-            max_tests: max_tests,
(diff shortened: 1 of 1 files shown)


BurntSushi:
@UnlawfulMonad The use of generators other than the thread local one should already be permitted using [`Quickcheck::gen`](https://docs.rs/quickcheck/0.6.0/quickcheck/struct.QuickCheck.html#method.gen), no?

That doesn't mean we shouldn't add `with_gen`, but I'd like clarification.

UnlawfulMonad:
@BurntSushi `QuickCheck::gen` uses the same type parameter `G` as an output so you can't actually use it to switch types (just to replace the generator with another of the same type). I can add a patch to make that method work as intended as well if you'd like.

BurntSushi:
@UnlawfulMonad Ah yeah I see, derp. That method is broken then. If fixing it isn't a breaking change (I don't think it is), then yeah, a patch in this PR would be great. We can keep `with_gen` too since it's a nice convenience and is consistent with conventions.

UnlawfulMonad:
@BurntSushi Should be ready to merge now after you review it. :)

fitzgen:
Ha! I had an item on my TODO list to fix this, finally got around to forking, and was surprised to see it was already fixed :)

Care to cut a new release when you have a moment, @BurntSushi ? Thanks kindly!

BurntSushi:
@fitzgen Ah yup, thanks for the reminder! `quickcheck 0.6.1` is now on crates.io.

PR #201, opened 2018-02-27
Upgrade to env_logger 0.5 and log 0.4; don't enable regex.

briansmith:
Upgrade to env_logger 0.5 and log 0.4 to reduce the need for
projects already using env_logger 0.5 and/or log 0.4 to compile
those libraries multiple times.

Don't enable the regex feature of env_logger by default, to avoid
forcing a regex dependency on dependent projects. A project can
enable the regex feature by adding a dependency on env_logger
without `default-features = false`.

Signed-off-by: Brian Smith <brian@briansmith.org>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,6 @@ i128 = ["rand/i128_support"]
 name = "quickcheck"
 
 [dependencies]
-env_logger = { version = "0.4", optional = true }
-log = { version = "0.3", optional = true }
+env_logger = { version = "0.5", default-features = false, optional = true }
+log = { version = "0.4", optional = true }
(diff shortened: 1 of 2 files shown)


BurntSushi:
Thanks @briansmith! I've merged this change with one tweak: I've kept `regex` enabled by default since it is also enabled by default in `env_logger`. To support your use case, I added a new `regex` feature, which should let you use `env_logger` with `quickcheck` without pulling in `regex`. e.g.,

```
quickcheck = { version = "0.6", default-features = false, features = ["use_logging"] }
```

`quickcheck 0.6.2` is now on crates.io with this change. Thanks again!

PR #203, opened 2018-05-22
Update rand crate to 0.5

oherrala:
This PR updates `rand` crate to version 0.5.

Minimum Rust version (in .travis.yml) pushed to 1.22.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.20.0
+  - 1.21.0
   - stable
   - beta
   - nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.21.0
+  - 1.22.0
   - stable
   - beta
   - nightly
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,4 +24,4 @@ name = "quickcheck"
 [dependencies]
 env_logger = { version = "0.5", default-features = false, optional = true }
 log = { version = "0.4", optional = true }
-rand = "0.4"
+rand = "0.5"
--- a/benches/tuples.rs
+++ b/benches/tuples.rs
@@ -14,7 +14,7 @@ macro_rules! bench_shrink {
             #[bench]
             fn $fn_name(b: &mut Bencher) {
                 // Use a deterministic generator to benchmark on the same data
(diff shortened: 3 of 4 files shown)


oherrala:
I'm not really comfortable with profiling Rust code on macOS yet, but running the `shrink_string` benchmark using Valgrind's callgrind I think I see a lot of `memmove()` from `VecShrinker`.

Looking at the code, there's `Arbitrary for String`'s

https://github.com/BurntSushi/quickcheck/blob/2143dc8ecfb01ab8fcf65e410243ba207ca23dc0/src/arbitrary.rs#L572-L576

which could produce quite a lot of `VecShrink`'ng.

How this is related to `rand` crate. I don't really know yet.

pitdicker:
I investigated the performance differences (hoping we didn't mess up too badly in Rand). I couldn't even find the RNG back when profiling!

The benchmark already notes "Use a deterministic generator to benchmark on the same data", and this is the trouble. The number of shrinkages (if I understand quickcheck right) depends on the range generated by the RNG. Any change and the benchmarks can become a magnitude faster or slower.

We did not change ISAAC, but did change the algorithm used in `gen_range`. It is easy to verify this is the problem: if you change the bechmarks to initialize the RNG with some other value, like `IsaacRng::new_from_u64(3)`, the results will be very different:

```
 shrink_string_1_tuple  64,943          12,420                 -52,523  -80.88%   x 5.23 
 shrink_string_2_tuple  90,180          24,018                 -66,162  -73.37%   x 3.75 
 shrink_string_3_tuple  190,919         64,927                -125,992  -65.99%   x 2.94 
 shrink_string_4_tuple  229,406         966,372                736,966  321.25%   x 0.24 
 shrink_string_5_tuple  296,096         1,094,523              798,427  269.65%   x 0.27 
 shrink_string_6_tuple  417,239         1,213,660              796,421  190.88%   x 0.34 
 shrink_string_7_tuple  1,029,913       1,280,004              250,091   24.28%   x 0.80 
 shrink_string_8_tuple  1,369,208       1,738,581              369,373   26.98%   x 0.79 
```

Now one is 5x faster, another 5x slower.

oherrala:
Sorry for taking so long to respond.

I agree with @pitdicker that randomness can affect the benchmark times. I also changed benchmarks to use Hc128 as @pitdicker suggested.

My feeling right now is that since benchmarks depend on how `rand` crate does things, it's not feasible to benchmark this PR and `master` branch.

PR #204, opened 2018-05-31
Re-export all of `rand`

jsgf:
Users of `quickcheck` need to use the same version of rand that quickcheck does if they want to make full use of rand's API, esp when implementing `Arbitrary`. This allows them to use `quickcheck::rand`, and still use another version of `rand` for other purposes if they wish.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,7 @@
 extern crate env_logger;
 #[cfg(feature = "use_logging")]
 #[macro_use] extern crate log;
-extern crate rand;
+pub extern crate rand;
 
 pub use arbitrary::{
     Arbitrary, Gen, StdGen,


jsgf:
See also https://github.com/rust-lang-nursery/api-guidelines/issues/170

BurntSushi:
Note that I am confused by two comments from you:

1. `rand` _isn't_ an internal implementation of quickcheck. `Rng` is part of the public API.
2. I don't know why you are drawing a distinction with respect to pre-1.0 crates. Could you elaborate on that?

jsgf:
1. You're correct - I phrased that badly. It's part of the public API, but it has the side-effect of also imposing quickcheck's version of rand on quickcheck's users.
2. If rand's API were stable, then being constrained to use the same version of rand as quickcheck is less of an issue. It's still a bit annoying, but less so (ie, I might not have bothered posting this). But as rand has an unstable API, being constrained to a specific version has a larger impact. But you're right - it doesn't have any fundamental effect on my argument, it applies either way.

jsgf:
> Do you know of any precedent with other crates that do this? For example, should every crate that derives serde::Deserialize (or serde::Serialize) also re-export all of serde?

That's a good question.

I hadn't run into problems with this before, but re-exporting seems like a generically good idea in retrospect, unless there's some specific reason one would want to constrain the API (and in that case, it's probably better to wrap the types rather than re-export).

It definitely depends on how the API is used. Maybe its not an issue with serde because the specific type or trait implementing the act of serialization doesn't matter, and they're polymorphic to the types that their users want so there's no type conflict. You'd need to do something like `Box<Serialize<X>>` to even notice that they're different `Serialize` traits.

In quickcheck's case, there's a lot closer coupling between Rng and Gen (and therefore Arbitrary), and Rng is fundamental to both quickcheck and rand's APIs, so you're stuck as soon as you want to implement Arbitrary for one of your types and want to go beyond Rng - which you must, since its only a trait.

So maybe this is a thing one would do on a case-by-case basis, but I think its definitely justified for quickcheck. I've had to patch it locally to unblock something I'm working on, so this isn't abstract for me.

jsgf:
https://users.rust-lang.org/t/rfc-api-guideline-re-export-a-whole-crate-if-exporting-any/17820

bluss:
It doesn't seem to be a problem to reexport rand, though I would be skeptical as well. (I have less time for Rust now! With that in mind it I gravitate more towards reexporting the crate just to solve this.)

Normally we'd just try to get everyone to use the new rand, gradually. I know that it bumps the Rust version requirement, is that the problem?

jsgf:
quickcheck going from 0.4 -> 0.5 would solve my immediate problem, but that will be a breaking change for all quickcheck's users.

BurntSushi:
@jsgf That's generally fine. I do my best to keep the pace of such changes infrequent, but they do happen from time to time. Getting to `0.7` over a period of four years feels about right. We should probably just move to 1.0 once rand goes to 1.0 too.

pitdicker:
Would it help to depend on `rand_core` in the signatures instead? https://github.com/BurntSushi/quickcheck/pull/203 will change the type bound of `Arbitrary` to `RngCore`. `rand_core` is intended to have very few (if any) breaking changes.

cc @dhardy

dhardy:
We are happy to add some kind of compatibility shim for Rand 0.4 → 0.5 if that would help; see https://github.com/rust-lang-nursery/rand/issues/421.

But as @pitdicker says, if you can get away with using only `rand_core` in the public API it will significantly reduce prevalence to breaking changes in new Rand releases.

BurntSushi:
OK, I've put `quickcheck 0.7` on crates.io. It no longer includes `rand` as a public dependency, but instead uses `rand_core`. Thanks for that suggestion!

I still didn't re-export the entire crate though. Let's see how this goes. If it continues to be a problem, we can revisit this.

PR #205, opened 2018-06-21
#[derive(Arbitrary)] V2

simonhdickson:
Follow up to #175, I believe I've fixed two of the issues reported in the other PR but I'm having trouble with generating nested tuples.

I'm quite new to procedural macros so if anyone could help me understand what I'm doing wrong I'd be grateful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ rust:
   - beta
   - nightly
 script:
-  - cargo build --verbose
-  - cargo test --verbose
-  - cargo doc
+  - cargo build --all --verbose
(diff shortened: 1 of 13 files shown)


simonhdickson:
Ok I solved my previous issue. Now when I expand the macro output it looks valid and when copied in place of the struct/derive it builds correctly, but when building normally it appears to have an issue saying `arbitrary expressions aren't allowed in patterns`.

Does anyone know how I can debug this issue?

BurntSushi:
https://github.com/BurntSushi/quickcheck/pull/222#issuecomment-573267560

PR #206, opened 2018-06-25
fix nightly, match header in ItemKind::Fn

PSeitz:
This fixes changes introduced in this commit https://github.com/rust-lang/rust/commit/56e8f29dbe89f2109cacc8eb5e92ea3de32eefb9


BurntSushi:
This is on crates.io in `quickcheck_macros 0.6.2`. Thanks!

PR #208, opened 2018-06-27
update dependency : rand="0.4" to rand="0.5"

bcko:
Thank you for your time checking this and I really appreciate your feedback&guidance! :)


bcko:
The Travis CI build failed for Rust 1.20.0

bcko:
@BurntSushi Is it possible to bump Rust minimum version to 1.22? If not, I will close this pull request.

BurntSushi:
I think this PR is a dupe of #203.

With respect to minimum Rust version, I don't think we have a choice. quickcheck is a heavy user of the rand crate, and we can't just not update. But yeah, this is annoying.

PR #214, opened 2018-09-20
Add support for Range{To}Inclusive

RReverser:


RReverser:
Resolved conflicts, let's see how it goes.

PR #219, opened 2018-11-06
macros: switch to using `proc-macro`

hcpl:
This makes `quickcheck_macros` usable on stable since Rust 1.30, but is a breaking change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ script:
   - cargo build --verbose
   - cargo test --verbose
   - cargo doc
-  - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
+  - if [ "$TRAVIS_RUST_VERSION" != "1.22.0" ]; then
       cargo build --verbose --manifest-path quickcheck_macros/Cargo.toml;
       cargo test --verbose --manifest-path quickcheck_macros/Cargo.toml;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.22.0
+  - 1.24.1
   - stable
   - beta
   - nightly
@@ -8,7 +8,7 @@ script:
   - cargo build --verbose
   - cargo test --verbose
   - cargo doc
-  - if [ "$TRAVIS_RUST_VERSION" != "1.22.0" ]; then
+  - if [ "$TRAVIS_RUST_VERSION" != "1.24.1" ]; then
       cargo build --verbose --manifest-path quickcheck_macros/Cargo.toml;
       cargo test --verbose --manifest-path quickcheck_macros/Cargo.toml;
     fi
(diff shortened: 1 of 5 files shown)


hcpl:
`lazy_static` bumped MSRV to 1.24 recently. What do we do?

BurntSushi:
Just bump Travis CI to Rust `1.24.1`.

Also, could you add a new Travis CI target for Rust `1.30.0` that tests the macro? We should also test it on beta as well. The `1.24.1` target will need to refrain from testing the macro.

hcpl:
> Just bump Travis CI to Rust `1.24.1`.

Done.

> Also, could you add a new Travis CI target for Rust `1.30.0` that tests the macro? We should also test it on beta as well. The `1.24.1` target will need to refrain from testing the macro.

There is an `if [ "$TRAVIS_RUST_VERSION" != "1.24.1" ]` before testing macros. Unless I misunderstood you, that should be enough?

hcpl:
Ping @BurntSushi as I still have unresolved questions in the comment above.

quickcheck_macros/Cargo.toml
```diff
@@ -14,8 +14,12 @@ autotests = false
...
-[dependencies.quickcheck]
-path = ".."
-version = "0.7.1"
+[dependencies]
+proc-macro2 = "0.4.19"
+quote = "0.6.8"
+syn = { version = "0.15", features = ["full"] }
+
+[dev-dependencies]
+quickcheck = { path = "..", version = "0.7.2" }
```
BurntSushi: Is it standard for macro crates to not depend directly on the crate they invoke?

.travis.yml
```diff
@@ -1,14 +1,14 @@
...
+  - 1.24.1
   - stable
   - beta
   - nightly
 script:
   - cargo build --verbose
   - cargo test --verbose
   - cargo doc
-  - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
+  - if [ "$TRAVIS_RUST_VERSION" != "1.24.1" ]; then
```
BurntSushi: Yup, this seems fine for now!

quickcheck_macros/Cargo.toml
hcpl: There are no API guidelines about this but that's how, for example, [serde_derive](https://github.com/serde-rs/serde/blob/a8a54c05681168d09dbdbb4212c31a142c3cdae4/serde_derive/Cargo.toml#L32), [failure_derive](https://github.com/rust-lang-nursery/failure/blob/8f8f92fa55e99d95a886cf54ad877f47059c7375/failure_derive/Cargo.toml#L18-L20) and [rocket_codegen](https://github.com/SergioBenitez/Rocket/blob/master/core/codegen/Cargo.toml#L28) behave.

The reasoning behind this is because macros only generate code, so they don't need the crate(s) they invoke to work. And depending on these crates won't gain any benefit --- users of the macros need to directly depend on those crates anyway.

ggriffiniii:
Just chiming in to say this would be nice to have. I would like to be able to use the proc macro attribute on stable rust.

Issue #190, opened 2017-11-30
Selecting values for usize


russel:
I have a test with signature:
```rust
fn a_test(n: usize) -> bool
```
but all the selected values of n are in (0, 100) which seems counter-intuitive. Is this what I should expect?

BurntSushi:
It is what the current implementation defaults to, yes. It is not ideal. A better generation strategy is welcome.

You can however change the range at which `usize` numbers are generated by [creating a `StdGen` with a custom size](https://docs.rs/quickcheck/0.5.0/quickcheck/struct.StdGen.html#method.new) and setting it on [`QuickCheck`](https://docs.rs/quickcheck/0.5.0/quickcheck/struct.QuickCheck.html#method.gen).

DurhamG:
@BurntSushi If we have to create a QuickCheck object and set a new generator on it, does this mean we won't be able to use the quickcheck! macro?

I have to admit, it's pretty absurd that a fuzz testing library only checks numbers <100.  We had a bug in some serialization/deserialization code of u64's and it was never invoked because it never received a number that required more than one byte to serialize.

BurntSushi:
> If we have to create a QuickCheck object and set a new generator on it, does this mean we won't be able to use the quickcheck! macro?

Yes.

> it's pretty absurd

You might reconsider your approach. This is unconstructive complaining that I absolutely refuse to deal with.

DurhamG:
You're right, it wasn't helpful.  Sorry, I was just frustrated.

I see someone proposed a different distribution [here](https://github.com/BurntSushi/quickcheck/issues/119#issuecomment-277247881).  Any thoughts on that style of solution? I can look into making a pull request if that approach sounds reasonable.

BurntSushi:
@DurhamG Ah yeah! That actually sounds like a great approach. We do the same thing for strings IIRC, so doing it for numbers makes a lot of sense. That's probably why I picked such a small range initially, to cover special cases that are typically small numbers (e.g., `0` and `1`). But moving that to a more sophisticated distribution that includes the full space of integers sounds great to me.

plul:
Do you have an update on this? #221 was closed without merging, but I cannot see why.

BurntSushi:
I didn't close it. Someone is welcome to resubmit it.

Issue #119, opened 2016-01-15
The random selection doesn't appear to be very random


sgrif:
I have a type which wraps an `i32`, and wrote an `Arbitrary` impl like so:

``` rust
trait Arbitrary<PgDate> {
    fn arbitrary<G: Gen>() -> Self {
          PgDate(i32::arbitrary())
    }
}
```

When I actually output what it's running with, the values are only ranging from `100` to `-100`, instead of 100 completely random numbers as I'd expect. Additionally, `shrink` appears to be flawed on these types. My understanding was that quickcheck would test for known problematic values for a given type. For `i32` I'd expect that to at minimum be `1`, `0`, `-1`, `i32::MAX` and `i32::MIN`, but when I add the `shrink` like so:

``` rust
fn shrink(&self) -> Box<Iterator<Item=Self>> {
    Box::new(self.0.shrink().map(PgDate))
```

I just see the same range of -100 to 100. This caused a critical bug that occurs for any values less than `-2451545` to go unnoticed.

BurntSushi:
@sgrif I think the original idea here is that the `size` method on `quickcheck::Gen` would control the magnitude of the integers generated, but perhaps that is a misuse. With #116 merged, I think there is now good precedent for biasing towards known problematic values.

vi:
It is somewhat reversed #99: numbers being biased too much towards `-100...100`.

I'm going to prepare a reformed number generation analogous to char (but simpler) which should cover entire range, yet preferring some particular numbers.

bluss:
There needs to be a new api for requesting a quickcheck-controlled size value. Something that's appropriate for selecting the size of a datastructure to generate. I think I mostly use usize's Arbitrary impl for that at the moment.

mulkieran:
Experienced a problem with some tests: https://github.com/stratis-storage/stratisd/pull/217/commits/0f7838d536619861899cb8d49106b750d59c787d, all u64 values were less than minimum required, so all testt were discarded.

bluss:
With current behaviour that is expected @mulkieran and you simply need a wrapper type that defines the distribution of values that you need.

mulkieran:
@bluss Can you give me an example?

mulkieran:
if anyone who understand this problem can give me some help, I'ld appreciate it.

bluss:
Something like this to make a wrapper type that puts a value inside using the regular `Rand` impl (not `Arbitrary`)

```rust
extern crate rand;
[...]

PR #221, opened 2018-11-30
Expand the ranges of arbitrary integers.

romanb:
This is supposed to address both #119 and #190 by interpreting the generator's size parameter as a bias towards the induced range, instead of as a hard limit. Specifically:

  * Values within the range induced by the size parameter occur with >  0.5 probability (currently ~.75 but it needn't be exposed to such a granularity on the API, so as to allow easy modifications / refinement).
  * Values outside of the induced range occur with < 0.5 probability.
  * Special values like 0, min_value and max_value generally occur  more frequently than any other value.

In accordance with intuition, lowering the size parameter increases the probability of smaller numbers and increasing the size parameter increases the probability of larger numbers being generated.

The distribution could be further refined but I was trying to keep it simple while solving the main problem of sampling from the entire range such that the influence of the generator's `size` is easily described and understood. [Here are some sample distributions with size=100](https://gist.github.com/romanb/ca0144b88e63cb79661ad64076004fc6) (the default size).

Further comments:

  1. Because I expanded the tests a little and also wanted to cover `i128` / `u128`, I bumped the dependencies on `rand` and `rand_core` to `0.6` and `0.3`, respectively, requiring a few minor modifications.
  
  2. Because the latest version of the transitive dependency `lazy_static` has a lower bound for rustc >= 1.24.1 I bumped the lower bound for quickcheck accordingly. If you are willing to go up to `1.26`, the feature gate for `i128` could be removed, as it stabilised in that release. Would that be desirable?

  3. The `assert!`s serve only to state the established invariants and could be removed or turned into comments, if desired.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -77,10 +77,14 @@ pub struct StdThreadGen(StdGen<rand::rngs::ThreadRng>);
 impl StdThreadGen {
     /// Returns a new thread-local RNG.
     ///
-    /// The `size` parameter controls the size of random values generated. For
-    /// example, it specifies the maximum length of a randomly generated vector
-    /// and also will specify the maximum magnitude of a randomly generated
-    /// number.
+    /// The `size` parameter is a hint for consumers of the RNG, i.e.
+    /// implementations of `Arbitrary`, as to the desired size of the generated
+    /// test data. The exact interpretation or influence of the size parameter is
+    /// up to the implementation of `arbitrary` for a type: Some may ignore it,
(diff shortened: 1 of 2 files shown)


BurntSushi:
@romanb Would you might rebasing this? I took care of the rand/rand_core upgrade and the MSRV tweak in separate commits.

romanb:
@BurntSushi Done. Since the MSRV is now 1.30, I could also remove the conditional compilation for the `i128` feature as part of this PR, if desired.

PR #222, opened 2018-12-07
#[derive(Arbitrary)] V3

aelred:
I wanted to fix the last issue in #205 so I could learn how to debug and write Rust macros.

With this fix, I think the `#[derive(Arbitrary)]` feature is complete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ rust:
   - beta
   - nightly
 script:
-  - cargo build --verbose
-  - cargo test --verbose
-  - cargo doc
+  - cargo build --all --verbose
+  - cargo test --all --verbose
+  - cargo doc --all
   - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
       cargo build --verbose --manifest-path quickcheck_macros/Cargo.toml;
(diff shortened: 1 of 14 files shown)


aelred:
It looks like the tests are failing on 1.22.0 because [`lazy_static` 1.2.0 requires 1.24.1](https://github.com/rust-lang-nursery/lazy-static.rs/releases/tag/1.2.0) - that dependency comes very indirectly from `env_logger`.

I suppose the choices are either to bump the minimum version to 1.24.1, or fix the `env_logger` version to an earlier one that doesn't use `lazy_static` 1.2.0.

maxbla:
@BurntSushi is this ready to merge?

BurntSushi:
I have neither reviewed this nor even decided whether this is something that should be in quickcheck, so no, it is not ready to merge. Personally, this is something I'm unlikely to want to maintain, and I don't think it needs to live in this repo.

BurntSushi:
I'm going to close this because I just don't have the bandwidth to maintain something like this at this time. I'd encourage folks willing to shoulder the maintenance to put this in a separate crate. Thanks!

Issue #187, opened 2017-11-03
Implement Arbitrary for [T: Arbitrary; N]


boustrophedon:
Obviously we can't do full generic N until const generics are implemented. I only wanted it for 3 `[f32; 2]`s anyway (and settled for six `f32`s in my function's parameters) and I'd be happy to send a pull request for just `[T: Arbitrary; k]` for a few small `k`s if that's good with you.

BurntSushi:
I think historically I've decided to not include these impls because of the compile time hit, with the hope that we could some day provide them with const generics.

I'm not convinced of this position btw. I suppose anyone that tries this should compare compile times.

jhpratt:
closed by #319

PR #226, opened 2019-01-10
Implement Arbitrary for arrays

### LINKED ISSUE #92: Add Arbitrary implementation for small arrays
<strike>Alternatively implement `Arbitrary` for all arrays using the `FixedSizeArray` trait from the standard library (requires nightly).</strike> (see below).

kurnevsky:
It increases compilation time from ~3.5s to 13.5s in release mode on my laptop. Also not sure how long shrinking can take for [T; 32]. What do you think?

Fixes #92 #187

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ default = ["regex", "use_logging"]
 unstable = []
 use_logging = ["log", "env_logger"]
 regex = ["env_logger/regex"]
+arrays = []
 
 [lib]
 name = "quickcheck"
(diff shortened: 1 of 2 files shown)


BurntSushi:
I think my previous opinion hasn't really changed, and a ~4x hit to compile times seemingly confirms it.

kurnevsky:
@BurntSushi I think it's essential to have this implementation to compete with proptest. If compilation time is so important we can have it behind a feature gate (added with the last commit).

BurntSushi:
I don't agree. We don't need to "compete" with anyone. If proptest is better for arrays, then use proptest.

BurntSushi:
We've made it this far without `Arbitrary` impls for arrays, we can wait a bit longer for const generics to land.

PR #234, opened 2019-05-04
impl Testable for Option<bool>

### LINKED ISSUE #105: Consider implementing Testable for Option<bool>
It seems like `Option<bool>` would be a simple way to discard tests vs. having to return a `TestResult` directly.

tobz1000:
Fixes #105.

Personally I think this is a worthwhile change for the ergonomics it enables. However, there is probably some ambiguity around the test result when returning `None`. Please let me know how best to document this if you think it's necessary.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -312,6 +312,16 @@ impl Testable for TestResult {
     fn result<G: Gen>(&self, _: &mut G) -> TestResult { self.clone() }
 }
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
I don't think my [previous opinion](https://github.com/BurntSushi/quickcheck/issues/105#issuecomment-175879063) has changed here. Sorry.

PR #235, opened 2019-05-11
Use `$crate::` prefix in `quickcheck!`

hcpl:
This makes possible using the macro as `quickcheck::quickcheck!` without `use quickcheck::quickcheck;` in Rust 2018.


BurntSushi:
Thanks! This fixed is on crates.io in `quickcheck 0.8.3`.

PR #236, opened 2019-06-07
impl<T: Arbitrary> Arbitrary for Bound<T>

cuviper:


cuviper:
Oh, the example fails on 1.30 because `RangeBounds::contains` was only added in 1.35. I guess I can mimic that functionality manually.

Issue #241, opened 2019-08-25
RFC: remove dependency on `rand` ecosystem


BurntSushi:
I am no longer happy about depending on the `rand` crates. There is too much churn, too many crates, and IMO, worst of all, there is no desire to [add a minimal version check to their CI](https://github.com/rust-random/rand/issues/741). Which means _anything_ that depends on `quickcheck` in turn cannot reliably have its own minimal version check.

Because I am tired of depending on `rand`, I have started removing it completely where possible. For example, in `walkdir`, I've [removed quickcheck as a dependency](https://github.com/BurntSushi/walkdir/commit/5d16163ce67245d07df0f913847b5f7a4c4c02b6). In ripgrep, I've [removed `tempfile` as a dependency](https://github.com/BurntSushi/ripgrep/commit/31807f805a793c254587105ca8ee0d41dfe3004b), because it in turn was the only thing bringing `rand` into ripgrep's dependency tree.

I don't see any other path forward here. I can either continue to grin and bear `rand`, drop everything that depends on randomness, or figure out how to generate randomness without `rand`. Specifically, I'd very much like to add a minimal version check back to the `regex` crate, which catches bugs that happen in practice. (See [here](https://github.com/rust-lang/regex/issues/570) and [here](https://github.com/rust-lang/regex/issues/593).) My sense is that there is some design space in the ecosystem for a simple source of randomness that doesn't need to be cryptographically secure, and an API that does not experience significant churn. Certainly, quickcheck does not need a cryptographic random number generator.

With that said, there is some infrastructure in the `rand` API that is incredibly useful. For example, quickcheck makes heavy use of the [`Rng::gen`](https://docs.rs/rand/0.7.0/rand/trait.Rng.html#method.gen) method for generating values based on type.

So it seems like if we have something like the `Rng` trait with with a non-cryptographic RNG, then we'd be probably good to go.

Are there other avenues here? What have I missed? My experience in building infrastructure for randomness is pretty limited, so am I underestimating the difficulty involved here?

Another side to this question is whether any users of quickcheck are leveraging parts of the `rand` ecosystem that would be difficult or impossible to do if we broke ties with `rand`.

elichai:
One thing you could do is have your own `Distribution` style trait that you implement for the types you want and then use one of the more minimalistic rand crates like `rand_os`

BurntSushi:
[...]

PR #242, opened 2019-08-27
Documentation added how to generate a struct

fspillner:
This is a PR addressing the missing documentation how to generate a struct, aiming for improving the "first-time" user experience with the library.

For that simple use case I needed to dig into the code of QuickCheck to find out, how to generate a arbitrary struct. And I had troubles with the missing Rnd instance for the given type, so I've added a notice about this. Maybe it's worth to have own topic for this issue.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# QuickCheck
+
 QuickCheck is a way to do property based testing using randomly generated
 input. This crate comes with the ability to randomly generate and shrink
 integers, floats, tuples, booleans, lists, strings, options and results.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -359,6 +361,42 @@ and the whole idea of property checking is to take that burden
...
+use quickcheck::{Arbitrary, Gen};
+// Important: Without this import the method gen() of Gen is not available then.
+use rand::Rng;
+
+impl Arbitrary for Point {
+    fn arbitrary<G: Gen>(g: &mut G) -> Point {
+        Point {
+            x: g.gen(),
+            y: g.gen(),
+        }
```
maxbla: This should be
``` rust
Point {
    x: i32::arbitrary(g),
    y: i32::arbitrary(g),
}
```

They way you did it depends on the `rand` api, which is prone to changing, see #241. Making my suggested change also obsoletes the accompanying notice (and the dependency hell that I think would come with it)

PR #244, opened 2019-10-06
[WIP] stateful property testing

Horusiath:
This is an initial PR with API for building stateful property-based tests using the `StateMachine<T>`. To bring the issue closer, here are the docs about the same feature implemented in [Erlang](https://propertesting.com/book_stateful_properties.html) and [F#](https://fscheck.github.io/FsCheck/StatefulTestingNew.html).

API so far was extended by two new structures:

- `Model` trait which defines 3 functions:
    - `run` which mutates current model's state given the generated operation
    - `next` which is supposed to produce new operation to be applied
    - `pre` which is supposed to check if a given operation has any sense in a context of a current model's state.
- `StateMachine<T: Model>` which is just a configurable wrapper around model, which implements `Testable` trait.

The minimal case for non-negative counter with `inc`/`dec` functions could look like this:

```rust 
#[derive(Clone, Debug, Default)]
struct Counter(u32);

impl Counter {
  fn inc(&mut self) -> u32 {
    self.0 += 1;
    self.0
  }

  fn dec(&mut self) -> u32 {
    self.0 -= 1; // better not to run below 0
    self.0
  }
}

// define test model

#[derive(Debug, Clone, PartialEq, Eq)]
enum CounterOp {
    Increment,
    Decrement,
}

impl Arbitrary for CounterOp {
    fn arbitrary<G: Gen>(g: &mut G) -> Self {
        if g.next_u32() % 2 == 0 {
            CounterOp::Increment
        } else {
            CounterOp::Decrement
        }
    }
}

#[derive(Default, Debug)]
struct CounterSpec {
    counter: Counter,
}

impl Model for CounterSpec {
    type Operation = CounterOp;

    fn pre(&self, op: &Self::Operation) -> bool {
        match op {
            CounterOp::Decrement => self.counter.0 > 0,
            _ => true // increments are always safe
        }
    }

    fn run(&mut self, op: &Self::Operation) -> bool {
        match op {
            CounterOp::Increment => {
                let expected = self.counter.0 + 1;
                expected == self.counter.inc()
            },
            CounterOp::Decrement => {
                let expected = self.counter.0 - 1;
                expected == self.counter.dec()
            }
        }
    }
}

#[test]
fn test_counter() {
    let spec = StateMachine::from(CounterSpec::default)
        .max_ops(50);
    QuickCheck::with_gen(StdGen::new(OsRng, 129))
        .quickcheck(spec);
}
```

At the moment, this library doesn't try to shrink the sequence of operations in case of failure.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,6 +85,7 @@ macro_rules! info {
 
 mod arbitrary;
 mod tester;
+mod statem;
 
 #[cfg(test)]
 mod tests;
--- a/src/statem.rs
+++ b/src/statem.rs
@@ -1,0 +1,204 @@
+
+use crate::tester::{Testable, TestResult};
+use crate::arbitrary::{Gen, Arbitrary};
+use std::fmt::{Debug};
+
+pub trait Model: Default + Debug {
+    type Operation: Arbitrary + Clone + Debug;
+
+    /// Generates a next operation given the current state of self.
+    /// Subsequently, this operation will be subject of [Self::pre] check to
+    /// determine, if it's a correct one in context of a current state.
+    fn next<G: Gen>(&self, g: &mut G) -> Option<Self::Operation> { 
+        Some(Self::Operation::arbitrary(g)) 
+    }
+
+    /// A preconditions used to check if generated operation is correct 
+    /// in sense of a current worflow. This function may be executed 
+    /// multiple times in a single run, therefore it's expected to not 
+    /// produce any side effects.
+    fn pre(&self, _: &Self::Operation) -> bool { true }
+
+    /// An actual operation to be run.
+    fn run(&mut self, op: &Self::Operation) -> bool;
+}
+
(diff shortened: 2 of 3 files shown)


BurntSushi:
Thanks for the PR!

Unfortunately, I'm quite sure what to do with this just yet. I don't really have the bandwidth to dig into this. Could you say a few sentences at a very high level about what this does and how you'd want it integrated into this library? Additionally, is this something that could live in a different crate?

Horusiath:
I'm not sure yet - I'd love to get some directions to get deeper insight into how shrinking actions works and how to integrate them into quickcheck. I'm quite tight on time myself, but would love to finish that contribution (especially given that it'd help me in my daily work).

BurntSushi:
All righty. I definitely don't have the time to mentor this, sorry. :( I'm going to close this for now, but if you happen to get this into another crate I would be happy to link to this in the README.

PR #248, opened 2019-12-30
Upgrade syn, quote, and proc-macro2 deps to 1.0

lopopolo:
Updated the version specifiers in `Cargo.toml` of `quickcheck_macros` and fixed the compiler errors.

I don't have any experience with `syn` and was following the docs and compiler errors. I think I got it right.


BurntSushi:
Thanks!

PR #257, opened 2020-02-06
Implement arbitrary for CString

### LINKED ISSUE #165: Implement Arbitrary for CString
Would be nice to have. `String`'s impl isn't quite the same.

ThomasdenH:
Closes #165. This implementation builds a `CString` from either a `String` or a `Vec<u8>`. It shrinks using the `VecShrinker`, everywhere making sure no null characters end up in the `CString`.

To prevent using as many allocations, it is built using an iterator, which is then filtered directly. I can change that if it's undesirable.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -3,8 +3,10 @@ use std::collections::{
     BTreeMap, BTreeSet, BinaryHeap, HashMap, HashSet, LinkedList, VecDeque,
 };
 use std::env;
+use std::ffi::CString;
 use std::ffi::OsString;
 use std::hash::{BuildHasher, Hash};
+use std::iter;
 use std::iter::{empty, once};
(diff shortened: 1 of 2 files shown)


maxbla:
@BurntSushi seems to keep pretty busy. He maintains quite a few libraries (just look at his github!). He takes a while to get around to reviewing PRs, but he often does eventually. Mentioning him with @ (as I already did) sometimes helps.

I can give you a comment in the meantime though.

I'm not convinced of the need for your `ArbitraryIterator`. You say it avoids excess allocations, but I don't understand how. You could do `(0..).map(|_| Arbitrary::arbitrary(g)).filter(|&c| c != '\0').take(size).collect()`, which I believe is equivalent to your CString code, both in final result and in number of allocations.

PR #265, opened 2020-12-27
remove rand as a public dependency + rollup and other cleanups

BurntSushi:
Part of this PR is meant to address #241, which was an unhappy reaction to how `rand` was evolving at the time. Since then, `rand` has slowed down its development and slimmed its dependency tree. It's still not quite as small as what quickcheck needs (which is really just a non-crypto PRNG), but it's closer.

One idea I and @matklad had was to just remove `rand` from the public API of QuickCheck. Doing this has a cost: it removes the ability to provide your own RNG to quickcheck and instead moves the use of the RNG to an implementation detail. My feeling is that providing one's own RNG to quickcheck was seldom-if-ever used, so I judged this to be an acceptable cost. In return, we are no longer pinned to `rand`'s release cycle _and_ we now have the freedom to switch out the RNG in the future if we want. For now, we upgrade to `rand 0.8` and otherwise leave it alone for now.

In addition to losing the ability to provide your own RNG, we also lose a lot of the lower level RNG convenience routines provided by `rand`. I believe we can capture much of that by just exposing concrete routines on `quickcheck::Gen`, in addition to the various `Arbitrary` impls.

The reasons why the `Gen` trait existed in the first place were:

1. I believe I copied the design as faithfully as possible from Haskell's QuickCheck, and IIRC, it is generic over the RNG.
2. At the time I wrote QuickCheck, RNGs were part of the standard library. Once RNGs moved out into the crate ecosystem, the costs of making that crate a public dependency manifested itself.

Anywho, I've written all this out so that it's clear what has changed and why. In particular, if you find yourself at this PR because some functionality in quickcheck has been removed, then **please open a new issue** that includes **the problem you're trying to solve** and we'll try to figure out how to move forward.


cuviper:
Coming here with an eye toward `rand v0.9`, I noticed that the README was never updated for this change:

> With all of that said, currently, rand is a public dependency of quickcheck. Therefore, the MSRV policy above only applies when it is more aggressive than rand's MSRV policy. Otherwise, quickcheck will defer to rand's MSRV policy.

(specifically the "public dependency" part)

Issue #268, opened 2021-01-11
`SignedShrinker` calls `abs()` on `i*::MIN_VALUE`, causing panic


jhpratt:
Consider the following snippet, using quickcheck 1.0.1.

```rust
quickcheck::quickcheck! {
    fn foo_can_shrink(v: i8) -> quickcheck::TestResult {
        let _ = quickcheck::Arbitrary::shrink(&dbg!(v));
        quickcheck::TestResult::passed()
    }
}
```

The test is trivial: it just ensures that some future value is smaller than the current.

When running `cargo test` on the snippet, I get (among other stuff)

```text
[demo.rs:3] v = -128
thread 'foo_can_shrink' panicked at 'attempt to negate with overflow', /home/jhpratt/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/num/mod.rs:98:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/panicking.rs:495:5
   1: core::panicking::panic_fmt
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/panicking.rs:50:5
   3: core::num::<impl i8>::abs
             at /home/jhpratt/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/num/int_macros.rs:1905:21
   4: <i8 as quickcheck::arbitrary::Arbitrary>::shrink::shrinker::SignedShrinker::new
             at /home/jhpratt/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-1.0.1/src/arbitrary.rs:817:40
   5: <i8 as quickcheck::arbitrary::Arbitrary>::shrink
             at /home/jhpratt/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-1.0.1/src/arbitrary.rs:860:21
   6: demo::foo_can_shrink::prop
             at ./demo.rs:3:17
   7: <fn(A) .> T as quickcheck::tester::Testable>::result::{{closure}}
[...]

PR #270, opened 2021-01-12
Eliminate panic on overflowing absolute value

jhpratt:
See #268. This ensures that the shrinker does not panic when _using_ the value, not just _constructing_ it.


BurntSushi (approved):
Thanks, nice catch!

BurntSushi:
This PR is on crates.io in `quickcheck 1.0.3`.

PR #278, opened 2021-02-09
api: add Gen::set_size and Gen::from_seed

### LINKED ISSUE #277: Gen from seed
Many crates implement `quickcheck::Arbitrary` and I was very happy that I could use these implementations in [dicetest](https://github.com/jakoschiko/dicetest) by implementing `quickcheck::Gen`. Unfortunately I can't do this anymore since quickcheck 1.0.

Would it be possible to provide a function that creates a `Gen` from a seed? Something like that:
```
impl Gen {
    pub fn from_seed(seed: u64, size: usize) -> Gen {
        Gen { rng: rand::rngs::SmallRng::seed_from_u64(seed), size: size }
    }
}
```

jakoschiko:
The seed still can't be set by QuickCheck users, but the new Gen
constructor is useful for other crates that use QuickCheck only for
its Arbitrary trait.

Closes #277

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -39,15 +39,33 @@ pub struct Gen {
 }
 
 impl Gen {
-    /// Returns a `Gen` with the given size configuration.
+    pub(crate) const DEFAULT_SIZE: usize = 100;
+
+    /// Returns a `Gen` with a random seed and the given size configuration.
+    pub fn new(size: usize) -> Gen {
+        Gen { rng: rand::rngs::SmallRng::from_entropy(), size: size }
+    }
+
+    /// Returns a `Gen` with the given seed and a default size configuration.
+    ///
+    /// Two `Gen`s created with the same seed will generate the same values. Though the values
+    /// may vary between QuickCheck releases.
+    pub fn from_seed(seed: u64) -> Gen {
+        Gen {
+            rng: rand::rngs::SmallRng::seed_from_u64(seed),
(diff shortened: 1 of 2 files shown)


jakoschiko:
@BurntSushi

Some questions:
- I called the new constructor `Gen::from_seed`. Is that okay? Or maybe `Gen::from_seed_u64`?
- You regret the current constructor `Gen::new` because it takes a size parameter. Should I add a `Default` implementation for `Gen` with random seed and default size?
- Are you sure that you want a `Gen::set_size`? `Arbitrary::arbitrary` takes a `&mut Gen` and this function makes it very easy to change not only the own local size, but also the size of the `Arbitrary::arbitrary` caller. Maybe we should provide a `Gen::with_size` that returns a copy.

audunska:
Just my thoughts here, but `Gen::with_size` seems reasonable here. Maybe even one that takes a closure? I.e., to generate arbitrary vec's with a size parameter of 10: 
`gen.with_size(10, Vec::arbitrary)`

It would be implemented something like
`impl Gen {
  fn with_size<T>(&mut self, size: usize, f: impl FnOnce(&mut Gen) -> T) {
    let old_size = self.size();
    self.set_size(size);
    let res = f(self);
    self.set_size(old_size);
    res
  }
}`

aakoshh:
What's the status of this PR? It would be super useful to have a seedable `Gen`.

JarredAllen:
^ I am also interested in this.

PR #283, opened 2021-03-26
Add Gen::gen_uniform

audunska:
Arbitrary impls for nested data often needs access to some kind of random numbers. To avoid exposing rand types, a simple uniformly distributed float on the unit interval is often sufficient, as many distributions of float and ints can be generated from this.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -76,6 +76,10 @@ impl Gen {
     {
         self.rng.gen_range(range)
     }
+
+    pub fn gen_uniform(&mut self) -> f32 {
+        self.gen()
+    }
 }
 
 /// Creates a shrinker with zero elements.
@@ -1582,5 +1586,15 @@ mod test {
                 PathBuf::from("/home/foo/../bar"),
             ],
         );
+    }
+
+    #[test]
+    fn gen_uniform() {
+        let mut g = Gen::new(10);
+        for _ in 0..100 {
+            let u = g.gen_uniform();
+            assert!(u >= 0.);
+            assert!(u < 1.);
+        }
     }
 }


src/arbitrary.rs
```diff
@@ -76,6 +76,11 @@ impl Gen {
     {
         self.rng.gen_range(range)
     }
+
+    /// Create a random number uniformly distributed on the unit interval [0, 1)
+    pub fn gen_uniform(&mut self) -> f32 {
```
nbraud, line 81: That should probably be `gen_uniform_f32` or somesuch, since there could as well be versions for `f64`, and all the integer types.

nbraud (changes requested):

src/arbitrary.rs
audunska: I guess one could make it polymorphic using num-traits..

BourgondAries:
Why can't we have `gen_range` instead to generate a uniform range? `0f32..=3f32` does not expand a rand type for instance.

BourgondAries:
Another note: Scaling up a uniform f32 `[0, 1)` range to `[1, 10)` via `x*9 + 1` leaves holes. There exists no float in `[0, 1)` that maps to `7.256350994110107421875`

akesling:
> Why can't we have `gen_range` instead to generate a uniform range? `0f32..=3f32` does not expose a rand type for instance.

Why not just make the existing `Gen::gen_range` public?  It's already used extensively in the provided `Arbitrary` implementations to good effect.

BurntSushi:
Because it introduces a public dependency on rand.

akesling:
> Because it introduces a public dependency on rand.

@BurntSushi Have you considered updating included implementations of Arbitrary only using the public interface?

Doing so should help clarify _by example_ how people should go about creating their own `impl`s.  Having "internal" implementations use `gen_range` and then not providing it publicly was... disappointing :(.

BurntSushi:
I've already acknowledged the deficiency elsewhere. I'm not ignorant of the problem. Ignorance isn't what is holding back improvement here. Time to dedicate toward maintenance is.

akesling:
> I've already acknowledged the deficiency elsewhere. I'm not ignorant of the problem. Ignorance isn't what is holding back improvement here. Time to dedicate toward maintenance is.

@BurntSushi apologies if I've offended.  I'm the ignorant participant in this thread just trying to understand the state of the world to use the library (and maybe help others do the same).  Do you have a link to discussion of changing the included implementations of Arbitrary?  The more cross-linked these discussions become, the harder it is for people like me to ask stupid questions.

Also, would you be open to a PR that migrates some Arbitrary `impl`s to only the public interface?

BurntSushi:
Again, the problem here is my available time to do maintenance. Organizing discussions is maintenance. And changing the Arbitrary impls to use only public APIs is not only impossible but a waste of my time.

I'm sorry but you'll have to look at recently opened issues and PRs to understand the state of the world here. There aren't that many.

BurntSushi:
I made a couple of edits to my last comment in case you're reading via email, FYI.

akesling:
For anyone who intersects with this discussion and needs the context I was lacking, start by reading https://github.com/BurntSushi/quickcheck/issues/241 and https://github.com/BurntSushi/quickcheck/pull/265 .

As it seems I've overstayed my welcome in this thread, I'll do my best to read a proper history of all of this (which I tried but apparently failed to do before posting initially).  Once there, I'll comment on whatever open PR/Issue is more pertinent, or, as per the request in https://github.com/BurntSushi/quickcheck/pull/265, open a more relevant issue.

@BurntSushi thank you for taking the time to help contextualize all of this.  I'm a fan of your work and can appreciate the burden of so much public code to manage.

PR #284, opened 2021-03-26
Shut up non_fmt_panic warning

neithernut:
Like a few other macros, `panic!` accepts a format string and arguments.
The format string is meant to be static and calling it with a message string will generate a warning.
Currently, the warning also indicates that `rustc` will generate an error instead in the near future.


neithernut:
Fixes #305.

schneiderfelipe:
@BurntSushi can we get this merged? ❤️

PR #289, opened 2021-05-17
Implement Arbitrary for fixed-sized arrays of Arbitrary elements

nbraud:
- [x] Implement `Arbitrary::gen`
- [ ] Implement `Arbitrary::shrink`

I tried implementing a non-trivial shrinker, but this ended up requiring `T: Clone`
and a lot of lifetime issues, so it seemed better to keep the dummy shrinker.

This PR depends on `const` generics, so that would bump the MSRV to 1.51 and
require a minor release.  Alternatively, this functionality could be hidden
behind an opt-in feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quickcheck"
-version = "1.0.3"  #:version
+version = "1.0.3" #:version
 authors = ["Andrew Gallant <jamslam@gmail.com>"]
 description = "Automatic property based testing with shrinking."
 documentation = "https://docs.rs/quickcheck"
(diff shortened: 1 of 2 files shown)


BurntSushi:
This looks like a duplicate of #282.

nbraud:
@BurntSushi Oops, yes, I missed that there was an open PR for it (apparently I only checked whether an issue existed...)
Sorry about that, feel free to close if the version in #282 is preferred.

nbraud:
I just saw the comment saying you'd rather not use `array-init`, closing.
Again, sorry for missing the existing PR  >_>'

PR #294, opened 2021-06-12
Do recursive shrinking without recursive function calls

### LINKED ISSUE #285: Stack overflow in quickcheck case shrinking
The following code causes `quickcheck` to stack overflow on trying to shrink the testcase.

```rust
/// Greatest common divisor of two integers.
pub fn gcd(a: i64, b: i64) -> i64 {
    assert!(!(a == i64::MIN && b == i64::MIN));
    if a == 0 {
        b.abs()
    } else {
        gcd(b % a, a)
    }
}


/// Returns (gcd(a, b), x, y) such that a*x + b*y = gcd(a, b) (ignoring overflow in the LHS).
pub fn egcd(a: i64, b: i64) -> (i64, i64, i64) {
    assert!(!(a == i64::MIN && b == i64::MIN));
    let mut r = (a, b);
    let mut s = (1, 0);
    let mut t = (0, 1);

    while r.1 != 0 {
        let q = r.0 / r.1;
        r = (r.1, r.0 - q*r.1);
        s = (s.1, s.0 - q*s.1);
        t = (t.1, t.0 - q*t.1);
    }

    if r.0 < 0 {
        (-r.0, -s.0, -t.0)
    } else {
        (r.0, s.0, t.0)
    }
}


#[cfg(test)]
mod tests {
    use super::*;
    use quickcheck::quickcheck;

    quickcheck! {
        fn qc_egcd(a: i64, b: i64) -> bool {
            if a == 0 || b == 0 {
                return true;
            }

            let (g, x, y) = egcd(a, b);
            g == gcd(a, b) && a.wrapping_mul(x).wrapping_add(b.wrapping_mul(y)) == g
        }
    }
}
```

neithernut:
We try to shrink values recursively, i.e. when a shrunk value witnesses a failure, we'd shrink that value further. Previously, this recursion would be implemented via actual control flow recursion, i.e. a function calling itself. Since the recursion could not be unrolled by the compiler, this could result in stack overflows in some situations.

Albeit such an overflow would often hint at a faulty shrinker (e.g. a shrinker yielding the original value), the stack overflow could also occur in other situations.

Fixed #285.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -344,7 +344,9 @@ impl<T: Testable,
             self_: fn($($name),*) -> T,
             a: ($($name,)*),
         ) -> Option<TestResult> {
-            for t in a.shrink() {
+            let mut r = Default::default();
+            let mut a = a.shrink();
+            while let Some(t) = a.next() {
                 let ($($name,)*) = t.clone();
                 let mut r_new = safe(move || {self_($($name),*)}).result(g);
                 if r_new.is_failure() {
@@ -353,16 +355,16 @@ impl<T: Testable,
                         r_new.arguments = debug_reprs(&[$($name),*]);
                     }
 
-                    // The shrunk value *does* witness a failure, so keep
-                    // trying to shrink it.
-                    let shrunk = shrink_failure(g, self_, t);
+                    // The shrunk value *does* witness a failure, so remember
+                    // it for now
+                    r = Some(r_new);
 
-                    // If we couldn't witness a failure on any shrunk value,
-                    // then return the failure we already have.
-                    return Some(shrunk.unwrap_or(r_new))
+                    // ... and switch over to that value, i.e. try to shrink
+                    // it further.
+                    a = t.shrink()
                 }
             }
-            None
+            r
         }
 
         let self_ = *self;


neithernut:
This change is not yet tested. Also, I'm not sure whether or not the `shrink_failure` function should be integrated into the final `match`. I suspect the primary reason to have a separate function in the first place was the recursion (although it doesn't say so explicitly in 5b19e7c21e1afdcabb894668ad61be70bb023ef0).

neithernut:
Sadly, I couldn't reproduce the failure using the reproducer from #285 (because the probability of hitting integer overflows is much higher). Hence, I used the following artificial test-case:

```rust
#[derive(Clone, Debug)]
struct Num(u32);

impl Arbitrary for Num {
    fn arbitrary(g: &mut Gen) -> Self {
        Num(u32::arbitrary(g) & 0x00ffffff) // Replace with contrete num if needed
    }

    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
        Box::new(self.0.checked_sub(1).map(Num).into_iter())
    }
}

fn fail(n: Num) -> bool {
    n.0 < 3
}

quickcheck(fail as fn (Num) -> bool)

```

While this would provoke a stack overflow on defde6fb0ce20b0c8c4e672aa9ae821f7d1f5b38 I'd get the expected `'[quickcheck] TEST FAILED. Arguments: (Num(3))` with these changes.

neithernut:
Just wanted to note: with these changes, faulty shrinkers which would previously provoke a stack-overflow may now produce endless repetition of a test for a small set of values (e.g. a single value).

It would thus be a good idea to bound the number of reproductions. We could introduce another setting to `Gen` which could be configured through an environment variable (e.g. `QUICKCHECK_REPRODUCTIONS`). I decided to hold off crafting such changes for now, as they would conflict with #287.

jonhoo:
Just came across a use-case where I hit the recursion limit (and where I actually don't _think_ a bug is involved), and this PR looks reasonable to me. @BurntSushi given your comment in https://github.com/BurntSushi/quickcheck/issues/285#issuecomment-830269194, it'd be useful to get your guidance on whether you think this is an acceptable path to take, or if not, what kind of shape you'd like to see an acceptable solution take.

Velnbur:
+, @BurntSushi encountered this issue too

PR #296, opened 2021-08-09
Make SignedShrinker bounded for <$ty>::MIN

### LINKED ISSUE #295: Infinite Repetition/Never Ending Test with `f32`
# Issue

I've got a test that uses the following structure:

```rust
quickcheck! {
    fn f32_quickcheck(f: f32) -> bool {
        if f.is_special() {
            true
        } else {
            let string = f.to_string();
            f == string.parse::<f32>()
        }
    }
```

The minimum failing input is:

```rust
f32::from_bits(0b11001111000000000000000000000000);   // -2147483600.0
```

Adding print debugging with `--nocapture` shows that this value, `-2147483600.0`, is called repeatedly, without ever stopping, if the checks pass. The actual implementation uses parsing for an arbitrary radix (using big-integer arithmetic), and a float formatter, but just checking pressing enter in the terminal shows that the quickcheck is completing each iteration faster than the newline can show.

A similar result also occurs for `f64` with the following structure:

```rust
quickcheck! {
    fn f64_quickcheck(f: f64) -> bool {
        if f.is_special() {
            true
        } else {
            let string = f.to_string();
            f == string.parse::<f64>()
        }
    }
```

The minimal failing input is:

```rust
f64::from_bits(0b1100001111100000000000000000000000000000000000000000000000000000) // f=-9223372036854776000.0
```

It's worth noting that these have a practically identical bit pattern, which may help debugging this issue.

The currently rustc/QuickCheck version is:

```bash
$ rustc --version
rustc 1.55.0-nightly (b41936b92 2021-07-20)
```

```toml
[[package]]
name = "quickcheck"
version = "1.0.3"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "588f6378e4dd99458b60ec275b4477add41ce4fa9f64dcba6f15adccb19b50d6"
dependencies = [
 "env_logger",
 "log",
 "rand 0.8.4",
]
```

# Debugging/Solution

Is there any way to print the seed so I can more reliably debug these issues? Although the same infinite loop is reproducing on each subsequent run, this is somewhat difficult to debug, given that it uses a random seed and never exits, making it very hard to submit a PR.

neithernut:
Previously, the `SignedShrinker` was guaranteed to yield a result if initialized with `<$ty>::MIN` (e.g. `i32::MIN` for shrinking `i32` and `f32`). This would result in endless shrinking if a test would happen to fail for these values.

Shrinking signed values is done by halfing a second value `i` and subtracting it from the original `x`. Thus `i` will be equal to `0` at some point. We choose to halt the iteration in this state, i.e. not yield any new values. Yielding `x - 0` would be pointless anyway, since that obviously equals the original value which was the initial witness for the test failure.

----

Closes #295.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -824,7 +824,9 @@ macro_rules! signed_shrinker {
             impl Iterator for SignedShrinker {
                 type Item = $ty;
                 fn next(&mut self) -> Option<$ty> {
-                    if self.x == <$ty>::MIN
+                    if self.i == 0 {
+                        None
+                    } else if self.x == <$ty>::MIN
                         || (self.x - self.i).abs() < self.x.abs()
                     {
                         let result = Some(self.x - self.i);


neithernut:
Maybe it makes sense to introduce shrinking tests which assert that the shrinkers included in this crate yield at most $number values?

PR #298, opened 2021-08-24
Implement the From<bool> trait for TestResult

schneiderfelipe:
Hi @BurntSushi, thank you for this very nice library!

This PR adds an implementation of `From<bool>` for `TestResult`.
It allows using `TestResult::from(true)` everywhere, and `true.into()` when the `TestResult` type can be inferred from the context, which I believe improves ergonomics in some situations:

```rust
// From the README
fn prop(xs: Vec<isize>) -> TestResult {
    if xs.len() != 1 {
        return TestResult::discard()
    }
    (xs == reverse(&xs)).into() // Instead of `TestResult::from_bool(xs == reverse(&xs))`
}
```

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -186,7 +186,7 @@ pub fn quickcheck<A: Testable>(f: A) {
 /// Describes the status of a single instance of a test.
 ///
 /// All testable things must be capable of producing a `TestResult`.
-#[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


schneiderfelipe:
@BurntSushi can we get this merged? ❤️

jhpratt:
GitHub is being buggy, so I've merged this manually 👍

PR #299, opened 2021-08-24
Avoid some warnings from Clippy

schneiderfelipe:
This does very small changes to remove some warnings Clippy complains about.

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -47,7 +47,7 @@ impl Gen {
     /// randomly generated number. (Unless that number is used to control the
     /// size of a data structure.)
     pub fn new(size: usize) -> Gen {
-        Gen { rng: rand::rngs::SmallRng::from_entropy(), size: size }
+        Gen { rng: rand::rngs::SmallRng::from_entropy(), size }
     }
 
     /// Returns the size configured with this generator.
@@ -282,8 +282,8 @@ impl<A: Arbitrary> VecShrinker<A> {
         };
         let size = seed.len();
         Box::new(VecShrinker {
-            seed: seed,
-            size: size,
+            seed,
+            size,
             offset: size,
             element_shrinker: es,
         })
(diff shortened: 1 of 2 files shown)


src/arbitrary.rs
```diff
@@ -1365,11 +1364,14 @@ mod test {
...
+                        panic!(
+                            "{}",
+                            format!(
+                                "Element {:?} was not found \
                              in shrink results {:?}",
-                            n, shrunk
-                        ));
+                                n, shrunk
+                            )
+                        );
```
neithernut, line 1374: This change doesn't look right. Why would you use a format-string in a `panic!` just to feed it the output of a `format!`?

neithernut (changes requested):
See #284 for alternative changes for the `non_fmt_panic`-bits. This is not to say that they can't/shouldn't be included here (too).

src/arbitrary.rs
BurntSushi: Yeah I think you can just drop the `format!` call.

src/tester.rs
```diff
@@ -162,7 +162,7 @@ impl QuickCheck {
 
         let n_tests_passed = match self.quicktest(f) {
             Ok(n_tests_passed) => n_tests_passed,
-            Err(result) => panic!(result.failed_msg()),
+            Err(result) => panic!("{}", result.failed_msg()),
```
BurntSushi, line 165: Hmmm I don't think this is right... What's the motivation for this change?

src/arbitrary.rs
```diff
@@ -131,9 +131,7 @@ pub trait Arbitrary: Clone + 'static {
 }
 
 impl Arbitrary for () {
-    fn arbitrary(_: &mut Gen) -> () {
-        ()
-    }
+    fn arbitrary(_: &mut Gen) {}
```
BurntSushi, line 134: I think I prefer how it was written since it's an impl for `()` specifically.

src/tester.rs
neithernut: Like with `write!`, `print!` or good old `printf`, passing dynamic string-data as format string (`panic!` can format messages for some time now) is frowned upon. Current versions of rust will report a warning during compilation for this line, and probably an error in the future.

BurntSushi: Ah I see. Gotya. Yeah this is good, thanks for catching it!

src/arbitrary.rs
schneiderfelipe: Oh, I see! I'll roll it back then.

neithernut:
Fixes #305.

src/tester.rs
neithernut: I'd be ok with this. But I doubt that we'll see anything merged in this repo in the near future.

jhpratt:
GitHub is being buggy, so I've merged this manually :+1:

PR #304, opened 2021-12-30
Only build debug representations of arguments on failing tests

### LINKED ISSUE #303: debug_reprs taking up 41% of test runtime
It seems like we're formatting the test case arguments even on passing test runs, e.g. here is the flamegraph of a test run where the test passed, but we see 41% of runtime spent building up `debug_reprs`.

image

Does this make sense to you? do you see a way of modifying the test runner to only render the debug representation on failure?

davidrusu:
fixes #303 

This cuts my prop test run-times nearly in half!

Here's the new flamegraph showing nearly 70% of time spent in test, previously this was 40%.

image
compare to prior run:
image

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -189,7 +189,7 @@ pub fn quickcheck<A: Testable>(f: A) {
 #[derive(Clone, Debug)]
 pub struct TestResult {
     status: Status,
-    arguments: Vec<String>,
+    arguments: Option<Vec<String>>,
     err: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


davidrusu:
@BurntSushi this change nets us a 50% reduction in quickcheck's overhead

It's a fairly small change, so hopefully quick to review. I'd love to see it merged in soon.

neithernut:
There are bug-fixes pending for months, so I wouldn't get my hopes up.

dan-da:
+1 for getting this merged asap.   else I'll need to start referencing @davidrusu's fork in our crates instead.

niyaznigmatullin:
+1 is there anyone who wants to maintain this crate, probably we ask @BurntSushi to give rights or just use some other fork, @davidrusu's for instance, to continue developing this. The same is with crates.io, we could just create another crate

PR #311, opened 2022-06-07
Add support for arrays powered by const generics without unsafe or libs

### LINKED ISSUE #187: Implement Arbitrary for [T: Arbitrary; N]

Mathspy:
Closes #187

Alternative to #282 that does not use `unsafe`

This seems to have been a heavily requested feature since 2017 now but the hold back was that implementing array support via macros heavily impacts compile times, is limited to some predefined arbitrary length and bloats the docs with unnecessary clutters. The wait was for const generic stabilization which is finally upon us, so we can utilize that and add array support!

Also lol while I was writing the shrink implementation I was like uff too much cloning there's probably a way to write this better that Sushi will tell me about, then while opening this PR I decided to double check there are no PRs open that already do this, which is when I found #282 and saw the exact same implementation I reached. Now I feel much less bad about the implementation

### DIFF
--- a/src/arbitrary.rs
+++ b/src/arbitrary.rs
@@ -248,6 +248,26 @@ impl_arb_for_tuples! {
     (H, 7),
 }
 
+impl<const N: usize, A: Arbitrary> Arbitrary for [A; N] {
+    fn arbitrary(g: &mut Gen) -> [A; N] {
+        [(); N].map(|_| A::arbitrary(g))
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = [A; N]>> {
+        let cloned = self.clone();
+        let iter = (0..N).flat_map(move |n| {
+            let cloned = cloned.clone();
+            cloned[n].shrink().map(move |shr_value| {
+                let mut result = cloned.clone();
+                result[n] = shr_value;
+                result
+            })
+        });
+
+        Box::new(iter)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


arkeet:
I was looking to add this as well and came up with exactly the same implementation before seeing there was already a PR. Would be nice to see this get in.

cloudhead:
FWIW I've published a fork of this library called `qcheck` under here: https://github.com/cloudhead/qcheck

It includes this change as well as a bunch of others that made sense to include.

jhpratt:
closing as #319 was merged

Mathspy:
Looks like the other PR doesn’t handle shrinking

jhpratt:
:facepalm: You're absolutely correct. Mind opening a PR for adding shrinking to the now-existing implementation?

Mathspy:
> 🤦 You're absolutely correct. Mind opening a PR for adding shrinking to the now-existing implementation? If you'd rather I can copy over your code myself.

Opened https://github.com/BurntSushi/quickcheck/pull/330

Thank you for taking over maintenance of the library!

PR #314, opened 2022-12-05
Fix simple example in README

lynn:
Maybe I'm doing something wrong, but the example didn't compile for me until I added a `use crate::reverse`:

```
PS C:\code\playground> cargo test
   Compiling playground v0.1.0 (C:\code\playground)
  --> src\main.rs:17:19
   |
17 |             xs == reverse(&reverse(&xs))
   |                   ^^^^^^^ not found in this scope
   |
help: consider importing this function
   |
15 |     use crate::reverse;
   |
```

I also ran rustfmt on it; the `vec![]` is a suggestion from it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Here's an example that tests a function that reverses a vector:
 extern crate quickcheck;
 
 fn reverse<T: Clone>(xs: &[T]) -> Vec<T> {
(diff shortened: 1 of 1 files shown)


jhpratt:
Superseded by a larger change on `master`.

PR #317, opened 2023-03-18
deps: update to syn 2.0

davidpdrsn:


quickcheck_macros/src/lib.rs
```diff
@@ -33,14 +33,21 @@ pub fn quickcheck(_args: TokenStream, input: TokenStream) -> TokenStream {
             if errors.is_empty() {
                 let attrs = mem::replace(&mut item_fn.attrs, Vec::new());
                 let name = &item_fn.sig.ident;
+                if let Some(variadic) = &item_fn.sig.variadic {
+                    // variadics are just for `extern fn`
+                    errors.push(syn::parse::Error::new(
+                        variadic.span(),
+                        "unsupported variadic",
+                    ));
+                }
```
davidpdrsn, line 42: In syn 2.0 `TypeBareFn` changed to use [`BareVariadic`](https://docs.rs/syn/2.0.0/syn/struct.BareVariadic.html) instead of [`Variadic`](https://docs.rs/syn/2.0.0/syn/struct.Variadic.html) used by [`Signature`](https://docs.rs/syn/2.0.0/syn/struct.Signature.html). I can't see any obvious way to convert between the two kinds of variadics but not sure quickcheck even works for variadic functions 🤔

jayvdb:
@BurntSushi , could you have a look at this please

PR #319, opened 2023-08-09
feat: support arbitrary arrays

aatifsyed:
Uses [`std::array::from_fn`](https://doc.rust-lang.org/stable/std/array/fn.from_fn.html), which was introduced in [Rust 1.63.0](https://blog.rust-lang.org/2022/08/11/Rust-1.63.0.html#stabilized-apis), ~1 year ago


aatifsyed:
@BurntSushi any interest in merging this?

PR #328, opened 2025-01-30
[WIP] Upgrade rand to 0.9

lopopolo:
Rand removed the `Uniform` impls for `usize` and `isize` and instead added [`UniformUsize`](https://docs.rs/rand/latest/rand/distr/uniform/struct.UniformUsize.html) and does not have a generator for `isize`. I've commented those bits out for now since this mucks with the design of the `gen` function quite a bit.


jhpratt:
Done on main.

PR #335, opened 2025-06-04
1.0.4

j-vanderstoep:
Publish the last 4 years of commits to crates.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quickcheck"
-version = "1.0.3"  #:version
+version = "1.0.4"  #:version
 authors = ["Andrew Gallant <jamslam@gmail.com>"]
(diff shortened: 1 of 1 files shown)


j-vanderstoep:
Would you be willing to run `cargo publish` on this?

I'm working on getting crates used by the Android project updated to rand 0.9. Reducing the number of versions of crates that we use is helpful in reducing disk space and memory usage on Android devices. Thanks!

jhpratt:
Hi there! Thanks for the impetus :)

I was brought on a few months back to work through the PRs and other things to be updated. I did so a little bit before getting pulled back to more important things, but quickcheck is almost at the top of my to-do list!

I'll defer to @BurntSushi on whether to cut a release now — he's the only one with publish permissions anyways. But know that there will be a fair amount of improvement coming soon!

PR #345, opened 2026-03-27
Clarified doc comment on TestResult::is_error

johnw42:


jhpratt:
This is correct as-is. Your "correction" wrongly includes compilation errors.

johnw42:
Being polite costs you nothing.

And I don't see how anyone reading a doc comment on the behavior of a method would think it refers to compilation errors. You can't call any method when there are compilation errors.

jhpratt:
> Being polite costs you nothing.

I stated a fact. Nothing was an attack on you personally.

PR #347, opened 2026-03-27
Added integration with anyhow

johnw42:


jhpratt:
Panicking with a non-string payload is incredibly uncommon and not worth even the slightest overhead.

PR #349, opened 2026-04-09
Fixed bug causing no arguments to be reported when no shrinking occurs

johnw42:
This is a fix for issue #344


johnw42:
Could someone take a look at this PR? It fixes a regression that can make debugging unit tests very annoying.

PR #350, opened 2026-04-09
Added eq and ne methods to TestResult

johnw42:
This change adds methods to `TestResult` to simplify getting failure messages that are comparable to those produced by `assert_eq!` and `assert_ne!`.


johnw42:
I'm withdrawing this PR because I think it makes more sense to just use assert_eq! and assert_ne! directly.

PR #352, opened 2026-08-24
macros: avoid trivial cast in #[quickcheck] expansion

### LINKED ISSUE #261: Trivial cast error when using `#[quickcheck]` inside macro
This is a minimal example:
```
crate::my_macro!(u32);

#[macro_export]
macro_rules! my_macro {
    ($my_type:ty) => {
        #[quickcheck]
        fn fuzz_my_type(my: $my_type) {
            // ...
        }
    }
}
```

Results in

```
error: trivial cast: `fn(u32) {fuzz_my_type::fuzz_my_type}` as `fn(u32)`
   --> ~/fuzz_tests.rs:13:9
    |
  2 |   crate::my_macro!(u32);
    |   ---------------------- in this macro invocation
... 
 13 | /         #[quickcheck]
 14 | |         fn fuzz_my_type(my: $my_type) {
    | |_______________________^
```

It works if I add `#[allow(trivial_casts)]`, but it should work without flags to the compiler.

rome-xi:
Closes #261.

`#[quickcheck]` expanded `fn_item as fn(..)` which trips `trivial_casts` when the attribute is applied inside `macro_rules` that substitutes a type. Coerce through a `let` binding instead of `as`, and add a regression test that denies `trivial_casts`.


hlop3z:
Verified the regression test gates the bug: with `lib.rs` reverted to the pre-fix `#name as #fn_type` expansion (keeping this PR's test and `[[test]]` entry), `cargo test -p quickcheck_macros` fails with the same `trivial cast` error as #261; with the PR applied it passes. (rustc 1.97.0)

One note for reviewers: the `Cargo.toml` addition isn't incidental — this crate sets `autotests = false`, so without the explicit `[[test]]` block, `tests/macro_type_subst.rs` compiles but silently never runs. Confirmed by removing it.

rome-xi:
Thanks for verifying the regression gate — appreciated. Happy to adjust further if BurntSushi wants a different shape for the test/macro change.

